PR #22, opened 2026-08-11
fix(ui): tighten key gaps, simplify Panels, replace lock badge

owenpkent:
## Summary

Three reported UI issues. All of them live in the shared QML, so they land in the C++ rewrite as well with no port.

### 1. Too much dead space between keys

The visible gap between two keycaps is the layout's `keySpacing` plus **twice** the inset each `KeyButton` leaves around its own background (mine + my neighbour's), and the inset was the larger contributor at 2 px per side.

- `KeyButton` background inset: 2 px to 1 px
- `keySpacing` coefficient: `0.0025` to `0.0018` of window width

Gap goes from 8 px to 5 px at a ~1700 px window. The reclaimed pixels go back into the keys: the inset gives 2 px directly, and the smaller gaps shrink the fixed-pixel overhead the sizer subtracts before dividing, so every keycap grows in both dimensions. Bigger targets on a mouse-driven keyboard. Navigation and Numpad follow automatically since they read `keySpacing` from the same property.

### 2. The Appearance menu was confusing

**The Number Row toggle is gone.** `showNumberRow` is now derived: true exactly when the active layout JSON carries no `number` row of its own. The full-size layouts have one; the compact variants do not, so the standalone panel fills in for precisely the layouts missing it. Digits are therefore always on screen in both views, and a full-size layout can never end up with a second narrower number row stacked on the one built into its JSON. It reads the layout rather than the `compactView` flag because an arrangement with no `-compact` variant silently falls back to full size.

**Compact View and the side panels are now exclusive.** The nav cluster and numpad cost roughly 470 px of width, which is precisely what compact exists to hand back, so having all three independent let you pick the small-screen mode and undo it one toggle later. Compact wins: both panels are forced off and their toggles render greyed with a one-line reason rather than disappearing.

The restore is the part worth reviewing. The forced-off state is **suspended, not saved**: `onShowNavigationChanged` / `onShowNumpadChanged` skip their `appSettings` write while `compactView` is true, so leaving compact reads the real preference back out, and `Component.onCompleted` applies the same rule on a cold start so quitting in compact does not resurrect the panels next launch.

Compact View also moved to the top of the Panels section, above the two toggles it gates, so the cause sits above the effect.

### 3. The right-click modifier lock rendered as a yellow blob

A locked modifier had three things stacked on it: the accent fill (locked implies active), a hardcoded `#ffd84d` 2 px ring, and a 15x15 gold rounded square holding a padlock emoji at `pixelSize: 9`. Two problems with the badge: at 9 px the padlock is a smudge, and Windows draws it through Segoe UI Emoji as a **colour** glyph, which ignores the `color` property outright. The gold was also hardcoded, so it fought all nine themes.

Since locked already implies active, the key carries the accent fill regardless, and the indicator only has to say "more than a one-shot" rather than "a different thing". It is now a solid 3 px bar along the bottom edge, on top of that same fill. Ring and badge deleted, `border.width` back to a plain 1.

Two details:

- The bar's colour comes from a new `_onFillColor` property, which is the luminance rule the label was already using inlined for its active/pressed state. Factoring it out means the label and the bar are the same ink and cannot drift apart, and it keeps both legible on the pale-accent themes (Blackboard, Spaceship) and on Typewriter, which is light outright. The label's computation is exactly equivalent to before, just deduplicated.
- The bar is inset horizontally past the corner radius rather than full-bleed, because `clip: true` clips to the bounding rect and not the rounded shape, so a full-width bar pokes past the curve.

Verified by rendering `KeyButton` headlessly in resting / sticky / locked states across four themes at both full and minimum (34 px) key size.

## Test plan

- `python check.py` green: ruff, ruff-format, mypy, full pytest.
- New `tests/test_qml_compact_view.py::TestCompactViewForbidsTheSidePanels`: force-off on enter, round-trip restore of the user's real preference, and both-off-stays-off.
- Tests that drove `showNumberRow` directly now drive `compactView`, since visibility is derived. `TestNumberRowPanel` gained a shown-in-compact case alongside the hidden-on-full-size one.
- `test_widest_row_drives_total_key_units` needed the panels explicitly off on both sides of its comparison: compact now drops the default Navigation panel, so its 3.0u was folding into the delta and the test was no longer measuring the main block.
- Existing `TestPanelsSitFlushWithTheGrid` / `TestEveryRowFitsTheContentArea` sweeps pass unchanged against the new `keySpacing`, which they read live rather than hardcoding.

Heads up: your venv was missing `hypothesis`, so `check.py`'s pytest leg was aborting at collection on the two property-test modules. Installed locally; worth adding to the dev requirements separately.

## Accessibility

Key targets get **larger** in both dimensions, which is the right direction for imprecise pointer input. No change to keystroke timing, repeat interval, or press/release feedback. The disabled toggles are dimmed with an explanation rather than hidden, so an unavailable setting reads as unavailable rather than missing.

## Not touched

No prediction engine, build/signing, or telemetry changes.


owenpkent:
## Security review: no findings

Reviewed the full branch diff for security implications. Nothing met the bar for a real vulnerability.

**Scope.** 13 files: 6 QML (reviewed in full), 5 documentation, 2 test modules. **No file under `src/` is touched by this branch.** Every subsystem in this project's threat model lives there and is untouched:

- Privacy / password mode and the `activeContextChanged` suppression (`keyboard_bridge.py`, `platform/password_detect.py`)
- Zip-slip import hardening (`PackManager.import_pack`, `data_export.py`)
- Telemetry consent (`telemetry.py`)
- `WS_EX_NOACTIVATE` focus flags (`keyboard_app.py`)

**Surfaces specifically considered and dismissed:**

- **Privacy-mode bypass via the newly always-on number row.** `showNumberRow` becomes derived, so the standalone digit panel now renders unconditionally in Compact View. Its keys route through the same `pressKey` / `pressKeyLiteral` bridge slots as every other key, and privacy suppression is enforced in the Python bridge rather than in QML. No new path reaches the prediction cache or the learning path, and no password character can bypass suppression that a `?123` digit could not already.
- **Panel-visibility logic as a privilege boundary.** The Compact View exclusion and the suspend-don't-save settings behaviour only gate which panels draw. They do not touch keystroke synthesis, modifier hold/release/lock state, or the `_context_buffer` / `_current_word` mirrors.
- **The disabled-toggle visual.** `SettingsToggle` gains only an `opacity` binding. `Item.enabled` already propagated to the underlying `MouseArea`, so the dimming is cosmetic and cannot produce a clickable-but-invisible control.
- **`KeyButton` changes.** The geometry constants and the `_onFillColor` luminance refactor are arithmetic over theme colours. QML property bindings are not `eval`, so there is no injection vector, and the lock indicator reads the same `isLocked` / `isActive` properties as before without adding bridge surface.
- **Swipe hit-test geometry.** Changed key sizing could in principle desynchronise the `charKeyRegistry` key centres, but `pushSwipeLayout()` is already driven off `onWidthChanged` / `onHeightChanged` and centres are read from live item positions. Drift here would be a correctness bug, not a security one.

No user input is parsed, no file or network I/O is performed, no subprocess is invoked, no deserialisation occurs, and no credential, token, or PII is handled anywhere in the changed code.

CI is green on all 7 checks, including OSV Scanner and both Socket Security checks.

owenpkent:
## Security / robustness audit of this branch

Fuzzed the new panel + compact-view state machine and the lock indicator. Two commits pushed here, plus one spun out to #23.

**Scope check.** The branch is QML, docs and tests only, so none of the security-critical Python (import hardening, data export, password detection, telemetry, key synthesis) is in the diff. The one path where branch data reaches Python is `savedLayout`, a persisted registry string that flows into a file load: `KeyboardBridge.setLayout` resolves it as an allow-list lookup (`layout_id in self._layouts`) against a pre-globbed dict, so the `-compact` suffix concatenation in `resolveLayoutId` cannot become path traversal. Clean.

**Found and fixed here.** Compact enforced the panel exclusion only at the transition, never after. Not reachable by a user today (the toggles are disabled, and `SettingsPanel.qml` with its ungated toggles is dead code), but the `settingChanged` dispatch was unguarded, so a preset, shortcut or restored panel would leave compact rendering the very panels it exists to remove. Guarded both branches.

**Found and spun out to #23.** The headless QML suite makes real HTTPS calls to the GitHub releases API three seconds after every load, and that thread emits into a bridge the fixture has already torn down — which took the pytest process down with an access violation once the fuzz suite made runs long enough to hit it. Merged in here so this branch stays green; it lands on `main` on its own.

**Left alone, flagged.** `KeyboardBridge.checkForUpdate`'s worker emits into `self` with no liveness check and `shutdown()` does not join it. Low practical risk since the bridge outlives the process, but it is the auto-update path, so it wants its own change per `CLAUDE.md`.

**What held up.** 40-step randomised sequences across 5 seeds (compact × panels × layout × width × layer) hold every invariant after each individual mutation. Hostile widths (0, -100000, 100000) clamp correctly, `keyW` never breaches its 30 px floor, no NaN. Restart persistence is correct: quitting in compact returns in compact with the panels off and the real preference intact, and the window width does not drift across three cold starts. The lock bar clears 3:1 against its fill on all nine themes and stays positive-width at minimum window width in both views.

`python check.py` green.

PR #27, opened 2026-08-14
docs: regenerate the screenshots, and script the way they are made

owenpkent:
## Summary

The six README screenshots predated Compact View, Snippets, the second symbol page and the current settings layout, so they showed a keyboard that no longer exists. All six are regenerated, and two features that had no picture at all get one: **Compact View** and the **Snippets** window.

The more useful half of this PR is that they are no longer taken by hand. `scripts/capture_screenshots.py` loads the real `qml/Main.qml` against a real `KeyboardBridge` under the offscreen platform plugin, drives it into each state, and grabs the window. The next refresh is one command, and two shots taken a year apart still frame the same thing.

## Three things in that script are load-bearing

- **The synthesizer is mocked.** A real `KeyboardBridge` types into whatever app has focus, and the script types to populate the prediction bar. Without the mock, running it sprays text into your editor.
- **`APPDATA` points at a sandbox.** That keeps the run off the real model, analytics and snippets. More importantly it keeps *my own vocabulary* out of a public README: the word cloud and the dashboard render learned words. The demo model is seeded from neutral prose for the same reason. The sandbox sits under `C:\Users\Public` rather than `%TEMP%`, because Settings → Data & Privacy prints the export folder on screen and a temp path carries the developer's Windows username.
- **Fonts are pointed at explicitly.** The offscreen plugin builds no font database on Windows. The first run produced a pixel-perfect layout with every single glyph rendered as a tofu box, which is easy to miss in a thumbnail.

One more trap worth recording: the theme id is `purple`, not `amethyst` (that is only the display name), and `themeData[id] || themeData["dark"]` falls back silently, so getting it wrong photographs as a second identical screenshot rather than an error.

## Also

The test count was quoted as three different numbers (641 in the badge, 842 twice, 1289 once). All four sites now read 1316.

## Test plan

- `python scripts/capture_screenshots.py` regenerates all eight PNGs; each was reviewed.
- `ruff check` / `ruff format --check` pass on the new script.
- No `src/` or `tests/` changes, so behaviour is untouched.


owenpkent:
Rebased onto #28 and regenerated: the screenshots now show the redrawn clear-context icon rather than the old glyph. Test count updated to 1341.

PR #28, opened 2026-08-14
fix: Tab clears the prediction context, and draw the clear-context icon

owenpkent:
Two reported issues, both in the suggestion bar.

## Tab should clear the context

Tab moves to the next field, so everything `_context_buffer` / `_current_word` / `_sentence_buffer` describe belongs to a field the caret has left. This is the same failure the outside-click signal exists for, arriving through the keyboard instead of the mouse: left alone, a pill tapped after tabbing inserts the tail of a word whose head is in the previous field.

Tab now runs `_reset_typing_context(keep_snippet_offer=True)`. It keeps a live snippet offer for the same reason the click path does — tabbing to the next field of a form is the likeliest thing to happen right after typing an email address, and it must not close the Save button.

**The word in progress is deliberately not learned first.** Tab is the accept-completion key in every IDE and shell, where `_current_word` is a *prefix* the app is about to finish, so learning it would feed the model `hel` every time the user completed `hello`.

It is now the fifth context-clearing signal, and the only one that works identically on every platform (the other four are Windows-only polls of the accessibility layer or the pointer).

## The clear-context icon was not centred

Correct, and it was not a subtle amount. The button rendered `⟲` (U+27F2) in a `Text` with `anchors.centerIn`, which centres the text *item* while the ink inside it sits wherever the font puts it. Measured on the shipped build, the ring the eye reads sat down-and-right of the circle it lives on, with the glyph's tail hanging outside it to the left.

It is drawn with a `Canvas` now — an arc plus a triangular head, both sized off the button's own width — so it is concentric by construction at every size and in every theme. Same lesson as the padlock badge that came off the keycaps: any glyph small enough to be an icon is at the mercy of the host font.

| before | after |
|---|---|
| ring off-centre, tail outside the circle | arc concentric with the button |

Two details in the drawing are load-bearing: the arc **stops short of the arrowhead**, because a stroke running under the triangle fills the notch that makes it read as an arrow at all; and the head is sized off the **radius**, not the line width, since a head scaled off a ~3 px stroke antialiases into a nub.

## The test for it is the interesting part

The obvious measurement is a trap. **The ink bounding box is centred in both versions to within half a pixel**, because the tail hanging off one side cancels the ring being pushed to the other — a bbox assertion would have passed against the bug it was written for. The test measures the *spread* of the ink about the button centre instead: a concentric ring puts every pixel at nearly the same radius, an off-centre mark smears across radii. Glyph scores 0.29, drawn arc 0.09, and the bar sits at 0.15 with both versions on the correct side of it.

## Test plan

- `python check.py` green: ruff, format, mypy (linux + win32), 1340 tests.
- Both new suites (`TestTabClearsContext`, `TestTheClearButtonIconIsConcentric`) were run against the unfixed tree and fail there, with the messages you would want.
- `TestTabClearsContext` includes the inverse half: the Tab keystroke must still reach the app (the reset must not swallow it), and an ordinary cursor key must not clear.

Screenshots in the README still show the old glyph; I'll regenerate them on #27 once this lands.


owenpkent:
Icon updated after review: the first pass left a quarter of the ring open, which put the arrowhead out on its own at the side and read as lopsided rather than as a circle with an arrow on it.

Now a nearly closed ring with a medium head (`r 0.29w`, stroke `0.062w`, head `0.62r`), picked from six candidates rendered by the real Canvas code at the actual 45 px button size. The concentricity test is unchanged and still passes; it also still fails against the old glyph.

PR #30, opened 2026-08-15
feat: predict numbers, phone numbers and email domains

owenpkent:
## Summary

- **The engine can now learn things that are not words.** `NgramPredictor._tokenize` splits on `[a-zA-Z']+`, so every digit and symbol was thrown away before a word reached the vocabulary. A phone number, a zip, a house number and an email address were in a blind spot the engine could not represent, and typing a digit *blanked* the suggestion bar outright. New `src/prediction/token_predictor.py` keeps those whole, matched by prefix.
- **Typing `@` offers email domains**, with the ones you have actually typed moving ahead of the built-in list.
- **Two things found while measuring the slow gate**, both worth having on their own: a duplicated SymSpell build in `HybridPredictor.__init__` that cost ~0.4 s on every launch, and the pre-push gate going from 25 minutes to under one.

## Related issue

No issue; requested directly.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Docs only
- [x] Build / CI / tooling
- [ ] Other (describe)

## Flagged per CLAUDE.md

**This changes the prediction engine.** Specifically it adds a source that is deliberately *not* part of the merge. The two are mutually exclusive rather than ranked: part-way through `owen@gm` or `555-123-` no English word is a plausible suggestion, so `KeyboardBridge._in_token_context()` picks a bar instead of blending sources. The existing n-gram / PPM / fuzzy merge is untouched.

**It adds a dev dependency** (`pytest-xdist==3.8.0`, pinned like the rest). No runtime dependency change; nothing new ships in the installer.

**It does not change the data-export schema.** The store rides in `ngram_model.json` under a `tokens` key, so it backs up, restores and clears with everything else and `SCHEMA_VERSION` stands. A model saved before this exists simply has no key, which loads as an empty store.

## What it will not remember, which matters more than what it will

Anything in this store is offered back in the suggestion bar, written to disk and carried in a Data Backup archive, so the admission rule (`text_patterns.is_learnable_token`, next to the other shape rules so "what does an email look like" has one answer) is deliberately asymmetric:

- Any digit run over **8 digits** that is not phone-shaped is refused outright. That puts card numbers, account numbers and a US SSN (in every written form) out of reach *by construction*, rather than by enumerating which identifiers are sensitive, which is a game you lose eventually.
- Phone numbers are admitted separately by the digit-**grouping** rule the snippet detector already uses, so `123-45-6789` is not mistaken for one.
- Entries are **re-validated on every load**, so tightening the rule later cleans an existing store instead of grandfathering things in.
- Selections go through a new `record_token_prediction_selected` that never receives the value, keeping it out of `word_freq`, the dashboard's "Top Words" and `analytics.json`.
- Nothing on this path logs token content, and the whole feature is inert in privacy mode.

`docs/PRIVACY.md` has a user-facing section.

## Test plan

- [x] `python check.py` passes locally (50.7 s: ruff 0.1, format 0.1, mypy 2.2, mypy-win32 2.2, pytest 46.2)
- [x] Added or updated tests
- [x] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

New coverage, each positive paired with the near-miss it must reject:

- `tests/test_token_predictor.py` (36): store behaviour, the SSN / card / bare-9-digit rejections, domain ranking, eviction, and the persistence round trip including re-validation of a hostile saved model.
- `tests/test_keyboard_bridge.py::TestStructuredTokenPredictions` / `TestEmailDomainSuggestions` (17): end to end through the keystroke path, including the regression itself (a digit no longer blanks the bar), that a typed SSN is not learned, that a mention still gets word predictions, and that backspacing inside an address keeps the domain bar.
- `tests/test_text_patterns.py::TestLearnableToken`: the admission rule directly.

The pairs that bite are the SSN family and ordinary prose containing a number, not the tidy examples: a rule like "7-15 digits with a separator" passes every tidy case and accepts an SSN, a card and an ISO date.

**Not yet driven in `python run.py`.** Worth a manual pass before merge, particularly the pill row width with domain pills showing.

## Accessibility check

No timing or precision changes. The point of the feature is fewer clicks: a ten-digit phone number is ten clicks, and the digit row is a layer hop away on Compact View.

## Notes for reviewers

**The pre-push gate.** It had reached 25+ minutes, which is the only way a gate fails: it gets skipped. The obvious fix is to run less of it, and measuring first found something better. Lint and both mypy passes cost ~5 s between them, so the gate *was* pytest, and pytest was slow for a reason unrelated to test count: building a `KeyboardBridge` cost ~1 s and the fixture builds a fresh one per test, ~1300 times. That is per-process setup, untouched by any amount of clever test selection and almost perfectly parallel. `pytest-xdist -n auto` takes the whole gate to ~50 s. **There is deliberately no fast-subset tier**, and `check.py`'s docstring says why so it is not re-litigated. `--install-hook` writes a `pre-push` hook so it runs on push rather than from memory.

**One test-side fix that will recur.** The four headless QML modules persist through a QML `Settings {}` element, which resolves to a process-*external* store (a key under `HKCU`). Three defined the same scope literal independently and the fourth imported it, so under `-n auto` several workers shared one scope and cleared each other's keys mid-test. It surfaced as `the window width drifted across restarts: [1160, 940, 1160]`, which is indistinguishable from the persistence bug those tests exist to catch. Now `tests/qt_settings_scope.py`, keyed by worker. **CI is deliberately left serial**: it is the release gate, its runners have far fewer cores, and the local loop was the problem.

**Worth a close look:**

- `pressPrediction`'s dispatch requires the pill to be in `_token_pill_inserts` **and** in `_predictions`. The second half is the safety property: a dozen sites in that file emit a pill row, so "clear the map at each one" would be another set of parallel blocks to keep in sync, and missing one leaves a stale entry tappable. Requiring the pill to be on the bar *right now* makes that impossible without touching any of them.
- Token pills bypass `_display_cased` on purpose. They insert verbatim, `_release_sticky_modifiers` drops a held Shift first, and the synthesizer already inverts its own shift for an OS Caps Lock, so the stored string is exactly what arrives. Running `_display_cased` over it would show `GMAIL.COM` under Caps Lock and then insert lowercase, which is the display/insert mismatch that function exists to prevent.
- The trailing-space rule (space mid-sentence, none after an email or phone) is the one piece of real judgement here. A login form that does not trim rejects `owen@gmail.com ` with a validation error the user then has to notice and undo, which costs far more than the single click a wanted space costs.
- `is_learnable_token` is where an over-eager future edit would do damage. The 8-digit ceiling looks arbitrary and is not: it sits below a bare 9-digit SSN while clearing every shape worth learning.

**C++ parity gap**, recorded in `docs/architecture/BACKEND_PARITY.md` footnote 9: this is Python-only. Nothing about it is platform-specific, so it ports wholesale. Port the admission rule *first* and against the same near-misses; it is the half that looks most droppable when transcribing and the half that keeps the feature from remembering what it should not.


owenpkent:
Second commit (67e399b) fixes a defect the first one caused: the new pre-push hook made `check.py` pop two blank console windows on `git push`. Every step launches a console-subsystem binary without capturing output so it streams to the terminal, which is correct when a terminal exists; run through `sh.exe` from a GUI-initiated push there is no console to inherit, so Windows allocates one per child. The two ruff steps are the fastest, so theirs are the ones you catch.

Children now get `CREATE_NO_WINDOW` only when there is genuinely no console. Note that is *two* checks, not the obvious one: `GetConsoleWindow` returns 0 for a process on a pseudo-console or with stdout on a pipe, both of which happen in ordinary terminals (measured returning 0 under a plain `python.exe` while investigating), so it also requires stdout to be a non-tty. That makes the hand-run path unreachable and keeps ruff's colour. Covered by the new `tests/test_check.py`.

Also worth recording a change that is deliberately **not** here. I tried memoising the SymSpell deletion index across instances, which halves `KeyboardBridge` construction (0.60s to 0.25s). It made no difference to the gate: pytest measured 46.2s without it and 49.7s with it, because under `-n auto` the wall clock is bounded by the slowest single chunk (one QML fuzz test is 16s), not by aggregate CPU. Reverted rather than kept, since it adds module-level shared mutable state to production code for a benefit that measurement says does not exist. Worker-count sweep for the record: n=8 gives 59.6s, n=32 gives 46-56s, so `-n auto` stays.

PR #35, opened 2026-08-16
fix(windows): restore always-on-top and the taskbar entry

owenpkent:
## Summary

Reported as "always on top isn't working". The keyboard was asking for it the one way that does not work.

`WS_EX_TOPMOST` and the topmost Z-order **band** are separate pieces of state. MSDN is explicit that the style is added and removed with `SetWindowPos`; `_apply_windows_extended_styles` wrote the bit into the style word with `SetWindowLongW` instead, which sets the first and leaves the second alone. The follow-up `SetWindowPos` then passed `SWP_NOZORDER` and a null `hWndInsertAfter`, explicitly asking the system to leave the Z-order exactly as it found it.

## Evidence

Measured on the running keyboard rather than reasoned about. Walking the real Z-order (`GetTopWindow` / `GW_HWNDNEXT`) while the reporter had it open:

```
  #  TOPMOST      title
  1  True         On-Screen Keyboard        <- Windows' own OSK
  2  True         MacroVox
  3  False        run.py - Quarry - Visual Studio Code
  ...
 15  True         Alpha-OSK                 <- us
```

`WS_EX_TOPMOST` reads **true** the whole time, and the window sits **fifteenth**, below a dozen ordinary windows including the editor it is supposed to float over.

Qt's `WindowStaysOnTopHint` does place the window in the band correctly, which is why this ever appeared to work at all. Writing the style word afterwards is what knocked it back out.

## The fix

- `WS_EX_TOPMOST` is gone from the `SetWindowLongW` write.
- `WS_EX_NOACTIVATE` **stays** there, because that one genuinely is settable that way, and it is what stops a key click stealing focus from the app being typed into.
- The `SetWindowPos` call now carries `HWND_TOPMOST` and no longer carries `SWP_NOZORDER`, so it does both jobs: enters the band, and re-reads the extended style via `SWP_FRAMECHANGED`.

The snippets window goes through the same function, so it is fixed along with it.

## Related issue

No issue: reported directly, with a screenshot.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs only
- [ ] Build / CI / tooling
- [ ] Other

## Test plan

- [x] `python check.py` passes locally (ruff, format, mypy under both platforms, full suite)
- [x] Added or updated tests
- [ ] Tested in `python run.py` (see below, needs the reporter's session)
- [x] Verified on Windows (Z-order measured on the live window)
- [ ] Verified on Linux
- [ ] Verified on macOS

Four tests fake the Win32 layer and pin both halves plus both inverses: the call targets `HWND_TOPMOST`, it does not pass `SWP_NOZORDER`, `NOACTIVATE` still goes through the style word, and `TOPMOST` no longer does. The Win32 layer is faked because this code is Windows-only and the suite runs on Linux too.

**The runtime check still wants a restart of the app by hand**, which is why that box is unticked. The diagnosis is measured, the fix is not yet observed working end to end.

## Notes for reviewers

**A trap for anyone verifying this by hand.** A non-elevated process cannot call `SetWindowPos` on an elevated window, so probing a keyboard started through `run.py`'s admin relaunch from an ordinary shell returns error 1400 (`ERROR_INVALID_WINDOW_HANDLE`) and proves nothing about the code. Reading the Z-order is not gated that way, which is why the evidence above is a Z-order walk rather than an attempted fix-in-place.

**If it recurs after a restart**, the remaining possibility is that something demotes the window later rather than it never entering the band, and the answer would be to re-assert on the foreground poll that already runs at 4 Hz. Deliberately not done here: there is no evidence for it yet, and this is the window the user depends on daily, so the change is kept to the root cause with nothing speculative attached.

## Accessibility check

This restores a property the tool is unusable without: a keyboard that sits behind the window you are typing into cannot be clicked.


owenpkent:
Second fix pushed onto this branch: the taskbar entry.

Reported separately (no taskbar button, so the minimise button had nowhere to go and clicking the pinned icon did nothing). Same class of defect as the first one, in the same function: the window styles we get are not the ones the code says we have.

`WS_EX_TOOLWINDOW` is what removes a window from the taskbar. Nothing in this repo sets it, and the comments have claimed for a long time that it 'was removed' precisely so minimise would have somewhere to go. **Qt adds it back on its own**, and the trigger is the order of operations: QML declares `visible: true`, so the window is already shown when `_apply_window_flags` calls `setFlags`, and applying a non-activating, frameless, always-on-top flag set to an already-shown window is the case where Qt decides it does not belong in the taskbar.

Measured:

| step | TOOLWINDOW | exstyle |
|---|---|---|
| shown, default flags | False | `0x00000100` |
| **setFlags on the shown window** | **True** | `0x08000088` |
| after clearing | False | `0x08040008` |

`0x08000088` matches the live keyboard's `0x08080088` bit for bit apart from the LAYERED bit window opacity adds. Applying the same flags *before* the first show does not set it, which is why every isolated test of the flag set looks correct.

The style is now actively cleared rather than merely not set, with `WS_EX_APPWINDOW` asserted alongside so the result does not depend on Qt leaving the rest of the word alone.

PR #74, opened 2026-09-05
feat: widen the full-size space bar and drop the symbol layer

owenpkent:
## Summary

- The full-size layouts carried one symbol page of 34 glyphs, reached from a `Sym` key at each end of the space row. All 34 are also in the Symbols & Emoji picker, one click away in the suggestion bar on every layout, so the row was spending its two widest keys after the space bar on a second route to a set that already had one. The page is removed.
- The 3.0u goes to the space bar: **6.0u to 9.0u**, half as wide again.
- Adds the CHANGELOG entry for the Symbols & Emoji picker, which shipped with none at all.

The two surfaces are genuinely different, which is why the page was built: a picker is for *browsing* (categories, paging, a Recent page, a window over the app you are typing into), a key is for *reaching* something whose position you already know. That distinction is real but not worth a fifth of the row a pointer visits after every word.

## Related issue

No issue. Follows a question about whether the `Sym` keys still earn their place now that the picker exists.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs only
- [ ] Build / CI / tooling
- [ ] Other (describe)

## Test plan

- [x] `python check.py` passes locally (ruff, format, mypy under both platforms, full suite in ~74 s)
- [x] Added or updated tests
- [ ] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

What was actually run, rather than what the boxes imply: the layout change is data plus QML, and `tests/test_qml_compact_view.py` loads the real `Main.qml` headlessly and fails on QML warnings, so a binding error cannot ship as a blank keyboard. On top of that I rendered `Main.qml` offscreen with the Navigation panel on and eyeballed the bottom row, which is how I confirmed the space bar reads correctly beside a 1.2u Ctrl and that the letter columns still line up. I did **not** run `python run.py` and put a real window on screen, so the last four boxes are honestly unchecked.

Tests:

- `TestFullSizeSymbolLayer` (data) and `TestTheFullSizeSymbolPage` (live QML) are deleted with the feature.
- `TestTheFullSizeSpaceRow` replaces them. Its load-bearing assertion is that the space bar's centre is still at 8.25u. I checked it is not vacuous in either direction: it fails against the pre-removal data on the width, and against a hypothetical asymmetric widening on the centre (9.75u against 8.25u).
- `TestTheLetterColumnsLineUp` is untouched and still passes, which is what pins Q over A.

## Accessibility check

This is a net reduction in required precision. The widening is symmetric **about the space bar's own centre**, not the row's: the two `Sym` keys flanked the row, so removing both and spending all 3.0u on space leaves the row at 14.6u with the bar centred exactly where it already was. No click that landed on the space bar before can miss now, and 1.5u of fresh target appears at each end.

The cost, stated plainly: Ctrl, Win and Alt move 1.5u outward on both sides, so bottom-row muscle memory for those five keys moves by a key width. That is the trade, taken deliberately, since the space bar is pressed after every word and those five are not.

## Notes for reviewers

**The dead branch in `Main.qml` is the one thing worth a second look.** The "a layer key whose target is already showing goes back to base" branch existed only for `Sym`, whose entry key sat on the always-visible space row and so had to be both the way in and the way out. I removed it rather than leaving it inert, and checked the premise rather than trusting the comment: every layer key in `qwerty-compact.json` targets a layer it is not on (`?123` from base, `=\<` from sym, `ABC` from sym and sym2), so the branch was unreachable for all of them. `symLayer` in the `isActive` switch went the same way; no compact layer key carries a `stateKey`.

**The `"layer": "base"` fields go too.** They were added to the three letter rows by the symbol-layer commit itself, not before it, so removing them returns the full-size layouts to declaring no layers at all. I verified this against `git show` of the parent commit rather than assuming.

**Deliberately not touched.** Del stays off the full-size grid and Enter stays at 2.3u; both were the other half of the same commit and are what puts Q over A. Compact View's `?123` and `=\<` pages stay, and are not removable by this argument: 13 units cannot hold letters and digits at once, so compact has no other route to either, and a picker is no substitute for a digit.

**The second commit is unrelated housekeeping**, and I would take it out if you would rather it went on its own. The Symbols & Emoji picker shipped with no CHANGELOG entry, so the next release would have gone out without mentioning a whole feature. The entry covers the picker only: the symbol layer landed in the same commit, is removed here, and never reached a tagged release, so there is nothing for a user to be told about it.

`CLAUDE.md` keeps a "Removed:" section in the same style as the swipe-typing one, so the reasoning is in the file rather than only in this PR.


owenpkent:
Pushed `02e64c8`: every full-size row is now flush at 15.5u, the "Flush and symmetric" option.

**Why this belongs here rather than on its own PR:** the space-bar widening below is one row of the same arithmetic, and the two together are what make the edges straight. Splitting them would mean landing a row that is flush against four that are not.

**What changed.** The five rows totalled 15.5 / 14.3 / 14.9 / 14.3 / 14.6 units, and `Main.qml` centres each one against the widest, so the left and right edges stepped in and out five times by up to 0.6u. Compact View has forbidden exactly this since it shipped (`test_every_row_is_exactly_13_units`); full size never got the rule.

The widths are **derived from each row's own middle-key budget, not chosen**, which is why they land where they do. Middles are 12.0 / 11.0 / 10.0u, leaving 3.5 / 4.5 / 5.5u for the outer keys. Top and bottom split evenly; home gives Caps whatever Tab has and Enter the remainder:

| | before | after |
|---|---|---|
| Tab, Caps, `\` | 1.3 / 1.6 / 1.0 | **1.75** each |
| Enter, both Shifts | 2.3 / 2.0 / 2.3 | **2.75** each |
| Ctrl, Win, Alt | 1.2 / 1.0 / 1.1 | 1.25 / 1.0 / 1.25 |
| space | 6.0 | **9.5** |

So the whole keyboard is 1.0u, 1.75u and 2.75u keys plus Backspace and the space bar.

**Both pinned invariants survive, and one got sturdier.** The space bar's centre is still 8.25u. On a flush row that is `(15.5 + left - right) / 2`, which collapses to `(15.5 + Win) / 2` as long as the four Ctrl/Alt keys are equal, so the rule to keep on any future retune is just: Win stays 1.0u and those four stay equal. That is now written down in both the test and `CLAUDE.md`.

W-over-S reduces to `inset_top + Tab == inset_home + Caps`. With the rows flush both insets are zero, so it is now simply **Tab and Caps being the same width**. Previously the differing indents happened to cancel the differing Tab and Caps widths, so the alignment held by a coincidence between four numbers, and any one moving broke it. It did break once, which is why Del is off the grid.

**One existing test had to change, and it is worth a look.** `test_the_space_row_still_costs_no_window_width` asserted `space_units < widest`. That was right while the space row had slack; under the flush rule it is tied with every other row, which is the intended state rather than a regression. The half that still bites is the gap count: `_widestRow` takes max units and max gap count *independently*, from whichever row has the most keys rather than from the widest, so a key added to the space row would cost a `keySpacing` even though its unit total cannot change. That is what it asserts now.

New coverage: `TestEveryFullSizeRowIsFlush` pins the rule, the two symmetric rows, and the shared-width consequence. `python check.py` green.

Rendered offscreen to check the result; all four edges line up and Q sits over A.

owenpkent:
Pushed `22aa656`. You were right, and the units fix was only half of it.

**A row is also as wide as its gaps.** It measures `units * keyW + (keys - 1) * keySpacing`, and the rows have very different key counts. Measured before this commit:

| row | keys | gaps | width | inset each side |
|---|---|---|---|---|
| number | 15 | 14 | 949.00 | 0 px |
| top | 14 | 13 | 947.00 | 1 px |
| home | 13 | 12 | 945.00 | 2 px |
| bottom | 12 | 11 | 943.00 | 3 px |
| space | 6 | 5 | 931.00 | **9 px** |

The space row is nine gaps short of the number row, so at the default window it was 18 px narrow and, being centred, sat 9 px inside the grid at each end. Compact View had the same shape at a smaller scale (10 to 12 gaps).

**Each row now absorbs its own shortfall into its own keys**, deriving `rowKeyW = keyW + (_widestRow.gaps - (keys - 1)) * keySpacing / rowUnits`. Two alternatives were rejected: widening each row's `spacing` to fill would have squared the edges too, and given the space row 5.6 px gutters against 2 px everywhere else; absorbing into the keys is invisible at about 1 px on a 60 px key. The widest row's shortfall is zero by construction, so `keyW`, the side panels and the window's width budget are untouched, which is why nothing else in the layout math had to move.

**It cannot be made exact, and the test says so.** Qt Quick's positioners snap child positions to whole pixels, so a row of fractionally-wide keys accumulates rounding and its centred origin can land a pixel either side of its neighbour's. Non-compact is exact; two of compact's four rows sit 1 px across. `TestEveryGridRowIsPixelFlush` asserts equal **widths** exactly and equal **origins** within a pixel, kept as separate assertions on purpose so a row that goes genuinely short fails loudly instead of hiding under the tolerance.

**One measurement trap, now written down.** A `Repeater` is itself a zero-sized `QQuickItem` sitting in the positioner beside its delegates, so a `Row`'s own `width` can carry a phantom pixel past the last key that draws nothing. Measuring the bounding box reports two compact rows as 1 px wider than they render; the helper measures first-key-left to last-key-right instead. That cost me a couple of wrong turns before I spotted it.

The letter columns survive: top and home differ by one gap, so q and a drift about 0.3 px, inside the half-gap residual `TestTheLetterColumnsLineUp` already documents. Pinned at a 1 px bound.

New tests were run against the pre-fix layout and fail there (9 px spread non-compact, 2 px compact). `python check.py` green.

owenpkent:
Reviewed. CI is green, and I checked the load-bearing claims against the data rather than taking the notes' word for them.

**The dead branch really is dead.** I walked every `type: "layer"` key in all four layout files against the layer its own row renders on. No key anywhere targets a layer it is already showing, the compact page-switch keys included, so the removed "a layer key whose target is already showing goes back to base" branch was unreachable for all of them and `symLayer` had no other reader.

**The rows are flush and the centre held.** All five full-size rows total 15.5u and all twelve compact rows 13.0u. The space row reads `ctrl 1.25, win 1.00, alt 1.25, space 9.50, alt 1.25, ctrl 1.25`, which puts the bar's centre at exactly 8.25u. The widening really is symmetric about the point the pointer already aims at.

**The per-row key width is algebraically right.** `rowKeyW = keyW + (widestGaps - (keys - 1)) * keySpacing / rowUnits` makes a row measure `rowUnits * keyW + widestGaps * keySpacing`, which is equal across rows exactly when the unit totals are, and leaves the widest row at `keyW` untouched. That is what keeps the side panels and the window's width budget unchanged, so the claim in the notes holds rather than merely sounding plausible.

**No residual references** to `symLayer`, `sym-top`, `sym-home` or `sym-bottom` anywhere outside the two docs that describe the removal.

One thing to fix before merge, in the description rather than the code:

## The body describes 2 of the 4 commits

It says the space bar goes 6.0u to 9.0u. After `02e64c8` and `22aa656` it is **9.5u**, and the flush-rows work is not mentioned at all, even though it is the larger of the two changes and it is what stopped `keyW` being the per-key width (the `rowKeyW` derivation above). Anyone reading this PR later, or lifting its summary into a changelog entry, gets the wrong number and misses the reason the grid delegate grew a per-row width at all.

Nothing else blocking from me.

---

Note on ordering: this conflicts with #75 and #76 on `CHANGELOG.md`, `CLAUDE.md`, `qml/Main.qml` and `tests/test_qml_compact_view.py`. This one is ready now, so it is the natural first to land, with the other two rebased onto it.

owenpkent:
## Security audit: clean

Audited as part of a sweep over the six open PRs. No findings. The product decision to drop the symbol layer was taken as given and not relitigated; what was checked is that the removal is complete and consistent.

- **No dangling references**: zero matches at the branch tip for `symLayer`, `sym-top`, `sym-home`, `sym-bottom` or the `Sym` display key across `Main.qml`, all three layout JSONs (`qwerty`, `dvorak`, `colemak`) and both rewritten test files. This is the one that matters most, since a dangling layer reference produces a QML binding error on `Main.qml`, and that ships as a blank keyboard rather than as a visible failure. The layer-switch branch is correctly simplified to `root.activeLayer = kd.target || "base"`, since the only key that needed the toggle-back behaviour was `Sym` itself.
- **`"literal": true` regression**: not reintroduced. The `sym-*` rows were the only rows carrying non-ASCII glyphs and were deleted wholesale rather than partially edited, so no key kept a non-ASCII glyph while losing its `literal` routing. That is the regression where `pressKey`'s `str.upper()` under Caps Lock turns the micro sign into a Greek capital Mu, emitting one glyph while the cap displays another.
- **QML `textFormat`**: no `Text` element added or moved. The `Main.qml` change is `KeyButton` width bindings plus the layer-branch simplification above.
- **Logging**: nothing new interpolating typed content.

`python -m pytest tests/test_layouts.py tests/test_qml_compact_view.py`: 153 passed, run against an isolated worktree of this branch.

PR #75, opened 2026-09-05
feat: add F13-F24 and make every function key programmable

owenpkent:
## Summary

- Adds **F13-F24** on their own optional panel. These are real keys every OS understands that almost nothing binds, which is exactly what makes them worth having: a game, OBS or a macro tool can take one without colliding with that program's own F5 or Alt+F4.
- Makes **any function key programmable** through a small action registry (`src/key_actions.py`): `hotkey` fires a chord in one click, `text` types a stored phrase, and `key` keeps the keystroke and only relabels the cap. That last one carries its weight: twelve identical F-keys with one bound inside Discord cannot be told apart on screen, so writing "Push to talk" on it is the difference between the key being usable and not.
- Adds a **Function Keys category in Settings** listing all twenty-four keys with what each one currently does; a tap opens the editor. Right-clicking a key on the keyboard still works and is quicker, but it cannot be the only way in: a dwell-click, switch-access, head- or eye-tracker and single-button adaptive pointer have no right button, and anyone using one could otherwise press a function key and never program one. The two row-visibility switches moved onto that page too, so one category owns the whole feature.
- **The function rows now fill the keyboard's width.** Each F-key used to be one grid column wide with the row centred, leaving an empty strip at both ends. The keys stretch instead, so an F-key is 75.6 px against the 58.7 px key below it at a 940 px window, about a quarter more target.

## Related issue

No issue.

## Type of change

- [x] New feature
- [x] Bug fix (see *Notes for reviewers* - two found while bringing the branch up to date, one pre-existing in the feature)

## Test plan

- [x] `python check.py` passes locally (ruff, format, mypy under both `--platform linux` and `--platform win32`, pytest)
- [x] Added or updated tests
- [x] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

New coverage: `tests/test_key_actions.py` (registry, sanitisers, store, persistence), `tests/test_qml_function_row.py` (which keys each row draws, keycap labels, the Settings list and its hand-off to the editor, the editor itself), and `TestExtraFunctionKeys` / `TestProgrammableFunctionKeys` in `tests/test_keyboard_bridge.py` (chord dispatch, modifier merge, sticky release, the text path, the key-scope guard). Every case that could fire wrongly is paired with the near-miss it must leave alone.

Both keyboard states and the new settings page were rendered headlessly and looked at, not just asserted on.

Platform keycodes are data-only additions: Windows VK 0x7C-0x87, macOS Carbon kVK_F13-kVK_F20. F21-F24 have no macOS virtual keycode and are deliberately absent rather than guessed, since an invented code would post some other key; a programmed action on those still works there, because that path never reaches the map. Linux needs no change, as X11 already names F13-F24 and `xdotool` takes them directly.

## Accessibility check

- [x] Works with default repeat-delay / repeat-interval settings (function keys do not repeat)
- [x] No new requirement for fast or precise input. The Settings list is the left-click route into the editor and its rows are far bigger targets than a keycap; the keys themselves gained about a quarter of their width
- [x] Visual feedback still works for low-vision users - a reassigned key takes the theme accent on the keyboard and an outline in the list, and a custom label drops one font step rather than overflowing the cap

### The two layout changes, which reverse decisions this repo had written down

Both were made after rendering the alternatives, which is what the existing notes asked for rather than forbade.

**The rows fill the grid width now, and the keys are what fill it.** `FunctionRow.qml` carried a note saying the centred one-column-per-key shape had won over three ways of filling the width, and not to revisit the inset "without rendering the result next to the number row". That is what was done: the options were rendered to scale against the real number row and compared, and stretching won on target width, which on a keyboard driven by an imprecise pointer outranks lining up with the column below. The accepted cost is that no F-key lines up with the key under it any more, and at 29% wider and 30% shorter the row reads a little bar-like. The rule survives pointing the other way, and says so in the file.

The second half is not a side effect: **the group gap is now fixed and the key width gives instead.** It used to be the other way round, and with the Edit toggle in the run this row was 13 keys against Compact's 13-unit grid, 3 px of slack, so the gap clamped to `keySpacing` and 4-4-4 rendered as one undifferentiated run on the view where telling twelve identical keys apart matters most. A gap that gives is a gap that disappears exactly when the row is tightest.

**The Edit toggle became a Settings page.** The toggle flipped both rows into an assign mode where a left-click opened the editor. The list answers the same reachability requirement strictly better: its rows are far bigger targets than a 36 px keycap; there is no mode to get into or out of, and the old mode's only exit was the same key that entered it, sitting one pixel from F12, so an overshoot there changed what every function key did next; and it is the only surface that shows an assignment the user has forgotten making, which twelve identical keycaps cannot. Removing it also took the row back to twelve keys.

**One piece of that is worth a look because it is a window-management hack.** Tapping a row hides the settings window before opening the editor. The editor is typed into with the OSK's own keys and the settings window cannot hold OS focus, so the editor cannot live inside it (the Deepgram key field carries the same note); leaving a 360x540 window parked mid-screen would cover the editor, the letter grid it is typed with, or both. `root.settingsReturnView` brings settings back on the same page afterwards, which is the one exception to "re-opening Settings always lands on the home grid" that `CLAUDE.md` documents. That rule is about a re-open; this is a return, and the home grid would lose the user's place in a list of twenty-four. The inverse is tested: an editor opened by right-clicking a key must not pop the settings window open behind it.

### The rebase, and what had changed underneath the branch

**This branch was 35 commits behind and has been rebased onto current `main`.** Three things had changed underneath it and are worth a look, because each was a silent failure rather than a conflict:

1. **Swipe typing was removed on `main`**, taking `registerCharKey` / `registerFn` and both key registries with it. The branch still wired them on the new panel and the new toggle, and had a test class asserting against registries that no longer exist. All of it is gone, along with the branch's addition to the deleted `SWIPE_TYPING.md`.
2. **The dead-space fix (#73) landed after this branch was cut.** `hitMarginH` / `hitMarginV` default to 0 and there is no cascade, so the new F13-F24 panel had a dead strip between every pair of keys. It is wired now, and the extra row is switched on in `TestNoDeadStripBetweenKeys::test_every_key_in_every_panel_takes_a_share_too`, which is where that concern lives post-swipe. Nothing had covered the new panel.
3. **`telemetry` became its own QML context property** while the branch was out. The new QML test module hand-rolled its context registration and so died on `ReferenceError: telemetry is not defined`; it now goes through `tests/qml_context.install_context_properties`, which is the eighth fixture that helper was written for.

### Three fixes beyond the merge

The last is a user-visible bug in the feature itself:

- `_send_text_action` hand-rolled the verbatim-insert prologue because `_commit_verbatim_insert` did not exist when it was written. It is now a one-line call to it. The copy was already one field behind (`_word_prefix_lost`), which is the parallel-blocks drift this repo warns about for sticky-modifier release.
- **Shift+arrow moved the caret instead of selecting** in the key editor. That is one of the three text-box behaviours a window which cannot hold OS focus has to supply itself, and the editor had the other two (clicks reaching the field, Tab changing field). Without it there is no way at all to select a range in a 500-character phrase with an imprecise pointer.
- **`setKeyAction` reported a no-op save as a failure** (own commit, `8a8b1e1`). `KeyActionStore.set` answers "did anything change", which is what decides whether the file is rewritten and the signal emitted; the slot forwarded that to QML, which asks "did my save stick" and flashes its failure toast on False. So opening the editor on a key that already did what you wanted and tapping Save showed a red "That key action could not be saved" over state that was exactly right.

### The rest

**No prediction-engine, build/signing or telemetry change.** `keyboard_bridge.py` is touched, but only to hoist `pressSpecialKey`'s key-name map to the class (a programmed chord resolves its action key through the same map, and a second copy would be one more pair of parallel blocks), to give `_send_key` an `extra_modifiers` argument, and to add the action slots.

**`key_actions.json` is deliberately not in the Data Backup archive.** Adding a fourth file to `_MODEL_FILES` means bumping `data_export.SCHEMA_VERSION` and writing the back-compatible import path, which this project requires alignment on first. Flagging it rather than deciding it here.

**Security-relevant surface, since every field is user-authored and a chord reaches the platform synthesiser:** chord modifiers and action keys are validated against **allow-lists**, never a deny-list; modifiers are stored in canonical order so two spellings of one chord cannot read as two; a chord key must be a known name or a single printable ASCII character, because the platform layers translate that range and have nothing to say about a control character or an emoji; text strips every C0 control character plus DEL, keeping newline and tab, the same rule as `snippets._clean_value`. Entries are re-validated on every load, so tightening a rule retroactively cleans an existing file.

The one new string that reaches a `Text` element is a key's user-authored label, in the Settings list; both `Text` elements that render it set `textFormat: Text.PlainText`, per the `AutoText` note in `GOTCHAS.md`.


owenpkent:
Reviewed. CI is green, and the design lands where this codebase's conventions point: one registry, dispatch through `KeyActionStore.execute` so the bridge never switches on an action type, allow-lists rather than deny-lists on every field that reaches the synthesiser, and a bool return QML actually honours.

Things I checked rather than assumed:

- **Windows keycodes.** VK `0x7C` to `0x87` for F13 to F24 is correct, and all twelve resolve to real scancodes through `MapVirtualKeyW` on this machine (`0x64` to `0x6E`, then `0x76` for F24). So they take `_make_vk_scancode_event`'s primary path rather than its `wVk` fallback, which is what makes them relay over the remote-desktop pipelines that scancode mode exists for.
- **macOS keycodes.** `kVK_F13` through `kVK_F20` match Carbon. F21 to F24 being absent rather than guessed is the right call, and the note saying why belongs where you put it.
- **Edit mode.** `pressSpecialKey`'s `_edit_mode_active` intercept returns before `_run_key_action`, so a programmed F13 cannot fire while the snippets editor or the prediction-edit popup owns the keystrokes.
- **No recursion.** A chord whose action key is another function key goes through `_send_key` straight to the synthesiser, never back around through `pressSpecialKey`.
- **The editor is a `Popup`,** so it correctly needs no `_wire_floating_windows` entry, and it carries `modal: false` plus `closePolicy: Popup.CloseOnEscape`. All ten of its `Text` elements set `textFormat`, and the field overlays set `mouse.accepted = false`, which is the snippets-editor lesson applied rather than re-learned the hard way.
- **Nothing logs a label or a payload.**

One fix:

## `KeyActionStore.save()` bypasses `src/atomic_write.py`

```python
with open(tmp, "w", encoding="utf-8") as fh:
    json.dump(payload, fh, ensure_ascii=False, indent=2)
tmp.replace(self._path)
```

CLAUDE.md is explicit that this is not how a store gets added here: *"A new persistent store must not write with a bare `open()`/`write_text`; route it through `atomic_write_text` or `atomic_write_json` instead."*

The rename is atomic, so the docstring's "tempfile then rename" is half true. What is missing is the flush and the fsync that `atomic_write_text` does before it renames, which is the part that decides whether the bytes are actually on the disk when the rename lands. Without them a crash or a power loss can leave a renamed but short file exactly where the good one used to be, which is the specific failure the module was written to close.

`src/snippets.py` is the store this module's own docstring says it follows ("saved synchronously on every mutation like `snippets.json`"), and its `save()` is:

```python
try:
    atomic_write_json(self._path, payload, indent=2, ensure_ascii=False)
except OSError as exc:
    _logger.warning("Failed to save snippets: %s", exc)
```

which is this whole method, minus the manual temp path and its unlink cleanup.

Nothing else blocking. Keeping `key_actions.json` out of the Data Backup archive and flagging it rather than bumping `SCHEMA_VERSION` unilaterally is the right way round.

---

Note on ordering: this conflicts with #74 and #76 on `CHANGELOG.md`, `CLAUDE.md`, `qml/Main.qml` and `tests/test_qml_compact_view.py`. It is the biggest of the three, so it is probably the one to rebase last.

owenpkent:
## Security audit: 1 MED. No arbitrary execution introduced.

Audited as part of a sweep over the six open PRs, and treated as the highest-risk of the batch on the assumption that "programmable" keys might mean user-defined actions that execute programs. They do not.

Only three action types were built: `key` (send the default keystroke), `hotkey` (fire a chord) and `text` (type a stored phrase). None touches `subprocess`, `os.system`, `os.startfile`, `QProcess`, `ShellExecute`, `eval`, `exec` or `__import__`, and a scan of every added line across all six open branches finds zero process-launching primitives. The `launch`, `macro`, `midi` and `layout` types named in the Modular Layouts design doc were not implemented here. `src/data_export.py` is untouched, so `key_actions.json` is not in the backup archive and none of this is reachable from an imported archive.

### MED: the new store does not use `atomic_write.py`

`src/key_actions.py`, `KeyActionStore.save`. It hand-rolls `open()` plus `json.dump()` plus `Path.replace()` rather than routing through `atomic_write_json`. Two consequences: there is no `flush()` / `fsync()` before the rename, so on some filesystems the rename can land before the bytes do, and the temp name is a fixed `key_actions.json.saving` rather than a `tempfile.mkstemp`-generated one in the same directory.

CLAUDE.md names this invariant directly: a new persistent store must not write with a bare `open()`, it goes through `atomic_write_text` or `atomic_write_json`. Real-world impact is modest, since the store holds only key bindings, but a crash or power loss mid-save can leave a truncated file that the tolerant loader then drops, silently losing the user's custom assignments.

### LOW, functional rather than security

`qml/components/KeyActionEditor.qml`, around the chord-capture branch. Escape is intercepted unconditionally to close the editor before the capture branch runs, so Escape can never be bound as a hotkey's action key even though `"escape"` is in `CHORD_SPECIAL_KEYS`. Noted only because it sits in the same code path as the chord sanitiser.

### Clean

- **Sanitisation**: `_clean_label` (collapse and cap at 12), `_clean_text` (strips all C0 controls except newline and tab, plus DEL, cap 500), `_clean_modifiers` and `_clean_chord_key` are all allow-lists rather than deny-lists. Adversarial tests exist for each: oversized file, corrupt file, per-entry drop without losing siblings, non-list modifiers, non-ASCII and multi-character chord keys.
- **Load caps**: `_MAX_FILE_BYTES` (256 KB) is checked via `stat()` before the file is opened, matching the `snippets.py` pattern. No separate entry cap is needed because assignments are structurally bounded to the 24 `FUNCTION_KEYS`. Malformed entries are skipped individually.
- **QML**: every new `Text` in `KeyActionEditor.qml` sets `textFormat: Text.PlainText`. The one place a raw user label reaches the screen, `FunctionRow.qml`'s `displayText`, flows through `KeyButton.qml`'s label `Text`, which already sets it. A 12-character label fits `` but cannot trigger AutoText HTML sniffing anywhere in this feature.
- **Key synthesis**: Windows `VK_F13` to `VK_F24` are `0x7C` to `0x87`, matching Microsoft's documented range with no collision against `VK_NUMLOCK` (0x90) or `VK_SCROLL` (0x91). The macOS Carbon keycodes for F13 to F20 match the known `kVK_*` values, and F21 to F24 are correctly omitted rather than guessed, since no macOS code exists. Linux needed no new mapping: keysyms `"F13"` to `"F24"` already pass through the existing generic fallback.
- **Typed text on Linux**: the `text` action reuses `_commit_verbatim_insert` / `send_text`, the same path `insertSnippet` and `insertGlyph` use, so the mandatory `--` end-of-options guard before arbitrary typed text is inherited rather than reimplemented. No new bypassable call site.
- **Edit mode**: `KeyActionEditor.qml` sets `setEditMode(true)` on open and `false` on close, routes input through `editKeyTyped` / `editSpecialPressed`, and uses `modal: false` with `closePolicy: Popup.CloseOnEscape` only, matching the prediction-edit popup and snippets editor. Both text fields use the `mouse.accepted = false` MouseArea pattern, so caret placement and drag-select still work.
- **Logging**: no new log line interpolates a label, an action payload or typed text.
- **Tests**: `tests/conftest.py` correctly adds `src.key_actions.get_config_dir` to the guard that keeps the suite off the real config dir.

`python -m pytest -k "function or fkey or macro or action"`: 103 passed. `tests/test_key_actions.py` plus `tests/test_qml_function_row.py`: 54 passed.

PR #76, opened 2026-09-05
fix: stop the Dictation settings clipping, and separate the arrow keys

owenpkent:
## Summary

Two reported UI defects, unrelated to each other, both from a screenshot.

- **The Dictation settings category ran off the right edge of the window.** The Deepgram paragraph was cut mid-sentence, the second model read "Nova 2, previous", and the language chips lost a word off the end of each line.
- **The arrow keys ran straight into the nav block above them**, so the nine keys read as one undifferentiated field and the arrows had to be found by reading the caps rather than by shape.

## Related issue

No issue, both reported directly with a screenshot.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs only
- [ ] Build / CI / tooling
- [ ] Other (describe)

## Test plan

- [x] Added or updated tests
- [ ] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

Stated honestly rather than by ticking boxes: I did not put a real window on screen. The arrow gutter was verified by rendering `Main.qml` offscreen and looking at the result, and both changes are covered by tests that load the real QML headlessly and fail without the fix.

## Notes for reviewers

**One control caused all three Dictation symptoms.** The model picker was a `Row`, which sizes to its content and never shrinks. Inside a `fillWidth` `ColumnLayout` that raises the whole column's implicit width, so the description paragraph and the language `Flow` were both stretched to a width the 360 px window does not have, and then clipped by it. It is now a `Flow { Layout.fillWidth: true }`, matching the language chips directly below it that were already correct.

**The test for it needs a caveat, and it is the interesting part of this PR.** The shipped model labels happen to fit at the font the headless environment resolves, so a sweep over the real data passes with the `Row` still in place: it cannot catch this regression, and a test built only on shipped data would have been one of the "cannot fail" tests this repo's conventions warn about. The guard therefore drives a deliberately over-long label, and was run against the pre-fix layout to confirm it fails there (worst right edge 470 px against a 360 px window). The general sweep over all six categories is kept, because the property is worth stating for future categories, but it is not what covers this bug.

There is also a test asserting that `flickArea.contentWidth` equals its own width. That is deliberately a documented trap rather than a signal: `contentWidth: width` is the binding, so the two are equal whether or not anything inside has overflowed, and anyone reaching for it as a proxy should find the assertion and the comment first.

**The arrow gutter opens above the cluster, not below it.** A `Grid` aligns a cell's content to the top, so making the Up row's cells taller on its own would have put the new space between Up and Left/Down/Right, splitting the cluster it is meant to set apart. The Up key is anchored to the bottom of its own cell instead. The key keeps its own height, so no hit area changes and the gutter is dead space by construction, which is the same trade `FunctionRow` already makes for its group spacing.

The nav panel is now the tallest element in the window, so the window is about 8 px taller. In exchange the side panels bottom-align with the main grid, which they did not before.

**One unrelated line**: `.gitignore` now covers `.claude/worktrees/`, a scratch checkout directory that should never be committed as this branch's content.


owenpkent:
## Blocking: the new test is red on both runners

`TestALongModelLabelIsContainedByFlow::test_no_sibling_is_dragged_wider_than_the_window` fails on CI while passing locally (run 33995516455):

- `Test (ubuntu-latest 3/4)`: 75.0px past the right edge (435.0px against a 360.0px window)
- `Test (windows-latest 3/4)`: 10.0px past the right edge (370.0px against a 360.0px window)

Its sibling `test_the_deepgram_paragraph_specifically_is_not_clipped` passes, and so does the general sweep. Only the containment assertion fails.

## Root cause: `Flow` wraps children, it never shrinks one

The Row to Flow change is right for the reported bug, which is several chips summing wider than the window. It does not cover the case this test actually drives, which is **a single chip wider than the flow**. The pill is

```qml
implicitWidth: modelPillText.implicitWidth + 20
```

with no upper bound, so one over-long label overflows whatever width the parent assigns it, `Layout.fillWidth` or not.

Measured on this branch with the test's own `LONG_MODELS`, under the fixed-advance placeholder font that reproduces the CI metrics:

```
window width: 360   tolerance: 2.0
max right edge: 760.0

  right=  760.0  w=  746.0  QQuickRectangle_QML_91
  right=  760.0  w=  746.0  QQuickMouseArea
  right=  750.0  w=  726.0  QQuickText   "A model with an extremely long descriptive la"
  right=  595.0  w=  581.0  QQuickRectangle_QML_91
  right=  595.0  w=  581.0  QQuickMouseArea
  right=  585.0  w=  561.0  QQuickText   "Another quite long descriptive model label he"
```

Every overflowing item is a model chip or a child of one. Nothing else in the category overflows, which is exactly why the paragraph test passes: no sibling is being dragged, the chips themselves are simply too wide. The property the test states is stronger than what `Flow` provides.

That also explains the odd 10px against 75px spread. This class deliberately does not take `requires_real_font`, unlike the sweep above it, so it runs under whatever font each runner resolves, and the synthetic labels land either side of the line depending on the metrics. Not gating is the right instinct here, since a synthetic label ought to be font-independent. The problem is that the fix is not, so the test is measuring the font.

## Fix, verified on this branch

Cap the pill at the flow's own width and elide:

```diff
                                     Flow {
+                                        id: modelFlow
                                         Layout.fillWidth: true
                                         spacing: 6
                                         visible: unifiedSettings.dictationModels.length > 0
@@
                                             Rectangle {
                                                 id: modelPill
-                                                implicitWidth: modelPillText.implicitWidth + 20
+                                                implicitWidth: Math.min(modelPillText.implicitWidth + 20,
+                                                                        modelFlow.width)
@@
                                                 Text {
                                                     id: modelPillText
                                                     anchors.centerIn: parent
+                                                    width: Math.min(implicitWidth, modelPill.width - 20)
+                                                    elide: Text.ElideRight
                                                     text: modelData.label
```

With that applied, the worst right edge goes from 760.0px to 360.0px under the placeholder font, and the module's ten tests still pass under real fonts. It makes the assertion true for any label at any metrics, rather than for the ones today's fonts happen to fit, which is what the test was reaching for in the first place.

Worth noting the trade it accepts: an over-long label now elides on the chip. That seems clearly right for a control that has to survive a 360px window, and it is the same choice the Microphone section immediately below already makes for long device names.

## While the file is open

The language-chip `Flow` directly beneath carries the identical unbounded shape:

```qml
implicitWidth: langPillText.implicitWidth + 20
```

Latent rather than live, since the shipped language labels are short, but it is the same three lines and the same failure waiting on a longer locale name.

## The rest reads well

The nav gutter is sound and the reasoning in the comment is the part I would have wanted written down: a `Grid` top-aligns its cell content, so growing the row-4 cells alone would have opened the space between Up and the Left/Down/Right row and split the very cluster it is meant to set apart. Anchoring Up to the bottom of its own cell puts it above them instead. The key keeps its own height, so no hit area moves and the gutter is dead by construction, which is the same trade `FunctionRow` already makes between its groups.

`test_content_width_is_not_a_usable_signal` is a good idea. Pinning a trap as an assertion so the next person reaching for `contentWidth` finds the reason first is better than the comment alone.

---

Note on ordering: this conflicts with #74 and #75 on `CHANGELOG.md` and `tests/test_qml_compact_view.py`. #74 is green and ready, so landing that first and rebasing this onto it is probably the least work.

owenpkent:
## Security audit: clean

Audited as part of a sweep over the six open PRs. No findings.

What was checked:

- **QML `textFormat`**: this diff adds and moves no `Text` element. The only QML change is `Row` to `Flow` with `Layout.fillWidth: true` on the dictation model chips in `UnifiedSettingsPanel.qml`, so no element could have lost a `textFormat: Text.PlainText` line in the move. That matters because `Text` defaults to `AutoText`, which sniffs for HTML and can fire an outbound request just from displaying an untrusted string.
- **Dictation API key**: the diff touches only the model picker's layout, nowhere near the key input or the `hasKey` section. No new binding or write path touches `dictationApiKey`, and nothing new is written into the Qt `Settings` layer, which on Windows is a plain registry key with none of the caps or DPAPI protection `dictation.json` has. Only `savedDictationEnabled` belongs in `appSettings`, and that is unchanged.
- **Logging**: no new `console.log` or log line interpolating typed content or a setting value.
- **Arrow-key hit margins**: `NavigationPanel.qml`'s Up `KeyButton` still receives both `hitMarginH` and `hitMarginV`. The new `arrowGap` (`keySpacing * 4`) is a deliberate dead gutter above the arrow cluster, the same trade already taken for `FunctionRow`'s group spacing, not a forgotten binding that would leave a silently dead strip. The new `TestTheArrowGutterSeparatesTheArrowsFromTheNavBlock` covers both the enlarged gap above Up and the unchanged ordinary gap below it.
- **`QtQuick.Layouts` in a grid-aligned panel**: not introduced. `NavigationPanel.qml` still imports only `QtQuick 2.15` and uses a plain `Grid`, so it cannot round each child up to a whole pixel and render wider than the grid it sits flush with.

PR #77, opened 2026-09-05
fix: learn acronyms the user teaches with per-letter capitals

owenpkent:
## Summary

- **"PR" and "PRs" could never be learned, however many times they were typed**, and neither could QML, SDK, npm or any other acronym. `NgramPredictor._is_plausible_word` rejects a 1- or 2-letter word that is not on the language profile's `short_words` list, and a longer word with no vowel. That is what keeps `th`, `sm` and `xqz` out, and by shape an acronym is the same thing, so `learn()` dropped both before the 3-sighting gate, `_link_context` formed no `a -> pr` edge across the gap, and the strip on load re-deleted any that somehow got in. No rule over the letters can separate the two cases.
- **The evidence that does is what the user paid to type it.** `learn_capitalization` only records an all-caps form when Caps Lock was off for the whole word, so two or more capitals in the stored form means shift or right-click on each letter individually. `is_taught_acronym` reads that, and it is the one exemption from the shape filter. No new store, no new setting: the capitalisation table already recorded exactly this, and already has the load caps, the backup archive and Clear Learned Data behind it.
- **Two capitals rather than one is the load-bearing half.** A single leading capital is what every word at a sentence start carries, so `Th` from an interrupted word reaches `learn_capitalization` exactly as an acronym does, and a one-capital rule would hand the fragment class this filter exists for a free pass.

## Related issue

No issue; reported directly ("it never learns PR or PRs, why").

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs only
- [ ] Build / CI / tooling
- [ ] Other (describe)

## Prediction engine (flagged per CLAUDE.md)

Three engine behaviours change, all gated on the same taught-acronym evidence and inert for anyone who never teaches one:

1. **`_is_plausible_word`** gains its first exemption, checked last so the common case is still one length test or one pass over the characters.
2. **`load()` merges `capitalization` before the fragment strip**, where it used to merge after. This is not tidying: the strip asks `_is_plausible_word`, so merging afterwards would delete every learned acronym on the way back in and the model could never hold one across a restart. `caps` is now normalised to a dict at the read, because the merge happening early means a malformed field would otherwise take the whole load down rather than just its tail.
3. **`get_capitalized` returns the taught form**, which is the only thing besides the "I" family it will capitalise. This is **not** the removed Tier 3 coming back, and two guards are what keep it apart: the word must not be in `_base_unigrams` (so `us`, `ok`, `it` keep their own casing however they were once typed) and the taught form must be acronym-shaped, all caps but for a plural `s` (so `ZigZaqCorp` stays lowercase). Tier 3 was wrong because it fired on ordinary words and on mixed-case forms typed lowercase; both are excluded by construction. It is needed because `_display_cased` can only mirror capitals the typed prefix is long enough to contain, and every capital in an acronym after the first falls outside one, so the pill read `Pr` and got retyped by hand.

`HybridPredictor._next_word_allowed` is the fourth site: both merge gates now call it instead of `_short_word_allowed` directly, so a taught acronym is allowed in the position it helps most. Letting `pr` into the vocabulary while the two-character gate still barred it from next-word pills would have been half a fix.

Measured against a temp model after teaching it four times:

```
before: "opened a " -> ['lot', 'little', 'bit', 'couple', 'moment']
after:  "opened a " -> ['PR', 'lot', 'little', 'bit', 'couple']
after:  "...two pr" -> ['PRs', 'probably', 'PR', 'proud', 'problem']
```

## Test plan

- [x] `python check.py` passes locally (ruff, format, mypy under both platforms, full pytest: 79.8s, all green)
- [x] Added or updated tests
- [ ] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

New coverage, every positive case paired with the near-miss it must still reject:

- `tests/test_ngram_predictor.py::TestTaughtAcronymsAreLearnable` covers the vocabulary half. The near-misses are a lowercase typing (teaches nothing), a one-capital form (`Th`), a Caps Lock typing, a shipped word (`us` keeps its casing) and a mixed-case brand (`ZigZaqCorp` keeps its casing). One case saves and reloads a model carrying both an acronym and a fragment, which is the regression guard for the merge-ordering change above.
- `tests/test_hybrid_predictor.py::TestTaughtAcronymsReachTheBar` covers the next-word gate end to end through `predict`, with an untaught fragment as the inverse.

The suite was run whole rather than only the touched files, because `TestCapitalization` and `TestShortWordsAreOfferedAsNextWords` pin the behaviour this change works around. `test_learn_capitalization_still_records_form` did fail on the first attempt, against a wider version of rule 3 that returned any taught multi-capital form: that is what the acronym-shape guard is for, and the test stands unchanged.

The app itself was not launched; the change is engine-side with no UI surface.

## Accessibility check

## Notes for reviewers

**The known limitation is the teaching gesture.** An acronym has to be typed once with per-letter shift or right-click; typed under Caps Lock it teaches nothing, because that is one click for the whole word and therefore says nothing about it. That is not a new precision requirement (right-click for the shifted character is how capitals are already typed here) but it is the feature's one cost, and it is stated in `CLAUDE.md` and the changelog entry rather than left to be discovered.

Two things deliberately not done. There is no allow-list of common acronyms, because the next missing one is always the point. And nothing was added to `language.ENGLISH.short_words`: that list is the answer to "is this short string an English word", which `pr` is not, so widening it would have made the profile carry something that is not true of the language.

`docs/architecture/PREDICTION_NOTES.md` and the per-algorithm docs are unchanged; the write-up lives in the new *Taught acronyms* section of `CLAUDE.md`, next to the two rules it modifies.


owenpkent:
## Security audit: 1 HIGH, worth fixing before this merges

Audited as part of a sweep over the six open PRs. The privacy gating is sound. There is one real crash bug on the keystroke path, reproduced live.

### HIGH: this PR turns dormant bad data into a keystroke-path crash

`src/prediction/ngram_predictor.py`, `load()` merges the persisted `capitalization` dict verbatim (`self.capitalization.update(caps)`) with no per-value type check. The new `is_taught_acronym` / `get_capitalized` / `_is_plausible_word` chain then does `for c in stored if c.isupper()` on whatever was stored.

**On `main` this was harmless.** `get_capitalized` documents that the table "is intentionally not consulted here", so a non-string value sat inert and nothing ever iterated it. This PR makes the table live again without adding value validation, which is what converts a latent data-quality issue into a crash.

Reproduced: a model file containing `{"unigrams": {"pr": 5}, "capitalization": {"pr": 12345}}` loads "successfully" (the internal exception is caught and logged), and then every later call to `is_taught_acronym("pr")`, `get_capitalized("pr")` or `_is_plausible_word("pr")` throws an uncaught `TypeError: 'int' object is not iterable` from `_update_predictions` / `learn()`. There is no surrounding `try` / `except` on that path.

**It is reachable from untrusted input.** `models/ngram_model.json` is in `_MODEL_FILES` in `src/data_export.py`, so it is replace-on-import from a Data Backup archive the user was handed. The import layer validates zip structure, member paths and sizes; it never validates the JSON *values* inside a model file. So a corrupt, hand-edited or crafted archive breaks prediction on every keystroke touching the poisoned word, in a tool somebody depends on to type.

It also breaks the project's own convention twice over. Malformed entries are supposed to be dropped individually rather than poisoning the file, and because the merge happens before the fragment-strip step, an exception there leaves `self.capitalization` already mutated with the bad data while the rest of `load()` keeps the predictor's prior state, so the load is no longer atomic on failure.

Fix is small: drop non-`str` values individually as the dict is merged, the same shape as the existing per-entry tolerance elsewhere in this loader.

### LOW: the new tests do not cover the case that actually crashes

The new tests are otherwise well built and genuinely falsifiable, with every positive case paired with its near-miss (caps-lock guard, one-capital guard, shipped-word guard, mixed-case-brand guard, save and load round trip). But none exercises a malformed individual entry. The `isinstance` guard this PR adds (`if not isinstance(caps, dict): caps = {}`) is tested only for "the whole field is the wrong shape", not for "one value is the wrong type", which is the case above.

### Clean

- **Privacy gating**: this PR adds no new write path into `learn_capitalization`. It touches only `ngram_predictor.py` and `hybrid_predictor.py` and reads data produced through the existing, unmodified, privacy-gated write paths. `src/keyboard_bridge.py` is untouched, and its two `learn_capitalization` call sites still gate on `not self._privacy_mode` with `allow_uppercase = not self._word_typed_under_caps_lock`.
- **Caps-lock guard**: unchanged, so all-caps typed under Caps Lock still cannot poison the table.
- **`_word_prefix_lost`**: still short-circuits learning for an interrupted word.
- **Logging**: no new log line interpolates typed content.

No privacy-mode negative test was added, which is correct here given there is no new write path.

`python -m pytest tests/test_keyboard_bridge.py tests/test_ngram_predictor.py tests/test_hybrid_predictor.py`: 607 passed.

PR #78, opened 2026-09-06
feat(prediction): seed the base context tables from a public language model

owenpkent:
## Summary

- The base context tables covered **260 bigram prefixes against an 18,989 word vocabulary**, so for about 98.6% of the words a user can type, next-word prediction fell back to raw unigram frequency. This generates them from a public language model instead: 13,437 bigram and 20,000 trigram contexts, worth **+1.8 / +1.7 points of keystroke savings and +5.8 / +5.2 points of next-word hit rate** on two held-out splits.
- Adds a real evaluation corpus. `ksr.py` measured against 30 sentences written by hand in this repo, which cannot resolve the one and two point differences it has been used to argue and gave no external reference for whether its answer was good.
- Documents the ARPA format, the licence rule that constrains which corpora we can ship, and a negative result the benchmark cannot see.

## Related issue

No issue. Follows the "even larger n-gram corpus from a real public source" gap in `docs/architecture/PREDICTION_NOTES.md`.

## Type of change

- [x] New feature
- [x] Build / CI / tooling (new generator script, new bench corpus)
- [x] Docs only (partly: three stale docs corrected)

## Prediction engine change (flagged per CLAUDE.md)

This touches the engine, so here is the detail.

**What changed in scoring.** One thing: the first word of a sentence is now conditioned on a `SENTENCE_START` row instead of falling through to `_top_unigrams_with_scores`. Everything else is data. The seed counts enter `self.bigrams` / `self.trigrams` through the same additive path `load_common_bigrams` already used, so `_context_probs`, the interpolation weights and the merge are untouched.

**Counts, not probabilities.** ARPA carries normalised log probabilities; our tables carry counts that `_context_probs` normalises itself. Each context is given the same total mass a single curated pair carries (50), split across its continuations in proportion to probability. The mass is constant per context deliberately: `_context_probs` weights user evidence as `U / (U + 5 + 0.02·B)`, so scaling `B` by model confidence would make the contexts the model is *most* sure about the easiest for one stray user typing to overturn. Reasoning in `docs/architecture/NGRAM_SEEDS.md`.

**Curated pairs still win.** The generated seeds load last and are additive, so a context in both accumulates and the hand-tuned pair stays anchored. Checked by eye: `how -> is, can, do, about` (ours) reads better for this app than the model's `how -> much, do, to, many`.

**Nothing user-facing changed.** No new setting, no new dependency, no format change to `ngram_model.json`, and the user half of the context split is untouched.

## Measured result

`scripts/bench/ksr.py --corpus aac-dev` / `aac-test`, cold start, five pills. Each row adds to the one above.

| | dev KSR | test KSR | dev next-word | test next-word | never predicted |
|---|---|---|---|---|---|
| curated seeds only | 47.3% | 48.7% | 22.9% | 25.2% | 15.0 / 14.3% |
| + `seed_bigrams.txt` | 48.4% | 49.8% | 26.3% | 28.4% | 13.8 / 13.3% |
| **+ `seed_trigrams.txt` (shipped)** | **49.1%** | **50.4%** | **28.7%** | **30.4%** | **13.6 / 13.0%** |
| + trigrams at 50k contexts | 49.1% | 50.7% | 28.9% | 31.1% | 13.6 / 13.0% |

Next-word hit rate is the number to read: it is what these tables address, and it moved four times as far as KSR, which averages it in with the mid-word completions the prefix beam already handled.

**Trigrams stop at 20,000 contexts** because 50,000 costs 2.5x the file and 67 ms more at launch for +0.0 on dev and +0.3 on test.

**Cost: 2.56 MB on disk, 121 ms at launch** (against 1.8 ms for the curated seeds alone), p50 predict latency 2.1 ms to 2.6 ms. Loading the unpruned 119 MB model instead would be 14.5 s and 518 MB, which is why the pruning exists.

## A negative result, kept visible

The sentence-start row is **worth exactly nothing on the benchmark**: 170 of 557 first words in the top five either way, byte-identical KSR, because the top five unigrams already contain the common openers and the metric only asks set membership.

It ships because the benchmark structurally cannot see the case it is for. Every line in the AAC sets is one sentence and normalisation strips the punctuation, so the second-sentence case never arises. That is where the engine used to condition the next word on the previous sentence's last word:

| context | before | after |
|---|---|---|
| `i am tired. ` | of, and, to, from, but | I, the, it, if, you |
| `thanks for the help. ` | me, you, with, to, the | I, the, it, if, you |
| `can you help me? ` | with, the, a, know, and | I, the, it, if, you |

Measuring it needs multi-sentence text with punctuation intact, which is a different corpus. Both the null result and the blind spot are written into `NGRAM_SEEDS.md` rather than papered over.

## Licensing

Everything added is **CC BY 4.0**, attributed in `THIRD_PARTY_NOTICES.md` and in each generated file's own header.

`NGRAM_SEEDS.md` records the rule this followed: shipping alongside MIT means public domain, CC0 or CC BY only. **CC BY-SA is out** (share-alike conflicts with MIT redistribution), **BY-NC is out** (MIT grants commercial use), **BY-ND is out** (pruned counts are a derivative). That rules out COCA, Wikipedia-derived lists, Leipzig, and the Enron and Norvig n-gram files. There is a tempting argument that frequency counts are uncopyrightable facts under *Feist*; the doc says explicitly not to build on it, because EU database rights cut the other way and MIT invites forks.

The evaluation sets live under `scripts/` rather than `data/` because the PyInstaller spec bundles `data/` wholesale and there is no reason to ship an eval set to users. The AAC **training** split is deliberately not included: a training set beside the test set invites seeding the model from it.

## Test plan

- [x] Added or updated tests
- [x] Tested in `python run.py` (not just frozen build)
- [x] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

`tests/test_seed_ngrams.py`, 13 tests, every positive case paired with the near-miss it must leave alone: the backoff section must not be read as counts, a 2-word context must not flatten into the bigram table, seeds must add to curated pairs rather than replace them, malformed rows skip one at a time, `<s>` must be unproducible by `_tokenize`, and sentence-start must fire at an empty context and after `.!?` while leaving mid-sentence and mid-word alone.

Startup was smoke-tested through the real `KeyboardBridge` rather than only the predictor: constructs in 0.77 s, and `''`, `'how '`, `'i am tired. '`, `'wheelchair '`, `'physical the'` all return sensible pills.

Not verified on Linux or macOS. The change is pure Python data loading with no platform surface, so the risk is low, but it is untested there.

## Accessibility check

Timing note: 121 ms is added to startup and about 0.5 ms to p50 predict latency. Neither is in the keystroke path in a way a user would feel, and the benefit is fewer clicks per word, which is the thing that matters most for slow motor input.

## Notes for reviewers

- **The backoff weights are generated and deliberately unread.** `load_seed_ngrams` skips the `\backoff:` section rather than paying load time for a table no caller uses. That stays known gap #3, and it is a *smaller* prize now: backoff rescues contexts absent from the tables, and this just filled 33,000 of them.
- **Three stale docs got corrected on the way** (first commit, reviewable on its own): `data/README.md` documented five built-in vocab packs that stopped shipping and pointed custom packs at the wrong directory; `proper_nouns.txt` claimed in its own header that the engine auto-capitalises from it, which stopped being true when the three-tier auto-cap was removed; `PREDICTION_NOTES.md` listed SymSpell as an unimplemented gap when `src/prediction/symspell.py` has been in the merge for a while.
- **`CLAUDE.md`'s benchmark figures now name their corpus.** 52.5 / 53.6 / 55.0% are all on the old 30-sentence set. They stay correct and comparable to each other, and are not comparable to anything measured from here.
- **`check.py` does not lint `scripts/`** (its ruff step is `src/ tests/`), so the new generator passes because I ran ruff and mypy on it by hand. Widening that is out of scope here but probably worth doing.
- The four commits are meant to be read in order and each stands alone.


owenpkent:
## Security audit: 2 findings, neither exploitable today

Audited as part of a sweep over the six open PRs. No exploitable vulnerability found. Two deviations from invariants this project sets for itself.

### MED: the new seed loader has no byte cap and no entry cap

`src/prediction/ngram_predictor.py`, `load_seed_ngrams`. Every other loader here caps its input: the main model loader has `_MAX_MODEL_FILE_BYTES` plus unigram, bigram-prefix and capitalisation entry caps, and the pack loader caps both bytes per file and entries per file. This one has neither, and no `stat()` before the open.

Not reachable from untrusted input today: the path is always the hardcoded `_DATA_DIR` constant in `hybrid_predictor.py`, never user-supplied and never import-controlled. Raising it because the file it reads is now ~159k generated lines rather than the 754 hand-written ones `load_common_bigrams` was built for, and it will grow with future regenerations.

### MED: shipped CC BY 4.0 data has no `THIRD_PARTY_NOTICES.md` entry

`data/seed_bigrams.txt` and `data/seed_trigrams.txt` are loaded at every launch and are derived from Vertanen and Kristensson's "Forum only language models" (CC BY 4.0). The attribution is correct in the generated files' own headers and in `docs/architecture/NGRAM_SEEDS.md`, but there is no entry in `THIRD_PARTY_NOTICES.md`. Only the benchmark eval corpus (`aac_dev.txt` / `aac_test.txt`) got one.

`data/README.md` states the rule: "Record the source of any new file in `THIRD_PARTY_NOTICES.md` at the same time you add it." Anyone auditing shipped third-party content via the canonical notices file would miss that the base prediction data itself is third-party CC BY material.

### Clean

- **ARPA converter** (`scripts/gen_seed_ngrams.py`): no `eval` / `exec`, no subprocess, no network fetch, no regex at all (so no catastrophic backtracking). It is an offline developer-run tool, never fed attacker input at build or runtime, and the source model is a documented manual one-time download rather than something the code fetches, so no hash-pinning gap of the kind `build/linux/build.py` handles for appimagetool applies.
- **Base / user separation**: `load_seed_ngrams` writes only to `self.bigrams` / `self.trigrams` and never references `_user_bigrams`, `_user_trigrams` or `_user_total`. The `bigrams[p][w] >= round(_user_bigrams[p][w])` invariant holds and the persisted-inflation bug fixed in the context-table split is not reintroduced.
- **Malformed lines** are skipped individually (per-row `try` / `except ValueError: continue` plus an arity check), covered by `test_malformed_rows_are_skipped_one_at_a_time`.
- **Logging**: neither the loader nor the converter logs typed or user content, only counts and paths.
- **Backup / pack surfaces**: the seed files are referenced by neither `src/data_export.py` nor `vocabulary_pack.py`, so they cannot be smuggled in or exported as user data.

### Coverage note

The shipped data was audited by pattern sweep, not line by line: targeted searches across all four data files for email shapes, URLs, long digit and phone-shaped runs, and a short profanity list, plus about 60 to 70 lines read by hand. One inherited item surfaced, `bullshit` collocations carried through from `data/google-20000-supplement.txt`, which the "no-swears" filtering already missed upstream. Pre-existing, not introduced here. A full slur-list pass over the 161k lines was not run.

PR #82, opened 2026-09-06
fix(security): verify the installer's digest once it is pinned

owenpkent:
Closes the gap I named at the bottom of #81. **Stacked on #81** (base is
`fix/updater-toctou`), so review that one first; GitHub will retarget
this to `main` when #81 merges.

## The gap

#81's pin makes the installer unswappable from before the signature
check until after the launch. But it is taken a moment *after*
`_download_with_cap` releases its own write handle, and that sliver is
still a window. Much smaller than the one the pin closed, since nothing
deliberately blocking runs inside it, but not zero.

Nothing can *prevent* a swap there without changing what the download
hands back, so this detects one instead.

## The fix

`_download_with_cap` now returns the SHA-256 of the bytes it actually
wrote rather than a bare bool, hashed as they stream. Hashing by
re-reading the file afterwards would be a second look at it, and the gap
between the two looks is the thing being closed. The caller re-computes
the digest **once the file is pinned** and aborts on a mismatch.

The ordering is what makes it sound: the comparison runs after the pin,
so the file is immutable while it happens and the check itself cannot be
raced. Reading by path is safe there for the same reason, since the pin
denies rename-over and delete, so the name cannot have come to mean a
different file. A swap inside the sliver is therefore always caught, and
forging a match needs a SHA-256 preimage.

It also catches a corrupt or truncated write, which is worth having on
its own: that previously reached the signature check and failed there
with a much less useful error.

Unlike the pin, this half is not Windows-specific.

## Contract change

`_download_with_cap` returns `Optional[str]` instead of `bool`. That
touches two stubs in the existing tests: the failure stub now reports
`None`, and `_fake_download` returns the digest of what it writes.

## Tests

The main one drives a swap into the exact sliver, by wrapping
`_pinned_for_execution` so it replaces the file before delegating to the
real pin, then asserts the install aborts and nothing is launched.

Paired with the inverse (an untampered download still installs), since a
check that rejected everything would satisfy the first test and silently
break auto-update for everyone. The digest contract itself is exercised
against the real `_download_with_cap` with a stubbed `urlopen`, so a
digest computed over the wrong bytes fails there rather than passing
everywhere by agreement with the stub.

Verified both halves fail against the unchecked code:

```
FAILED test_a_swap_inside_the_gap_aborts_the_install
FAILED test_a_truncated_file_is_caught_too
```

`python check.py` passes.

### DIFF
--- a/src/updater.py
+++ b/src/updater.py
@@ -64,6 +64,7 @@ from __future__ import annotations
 
 import contextlib
 import ctypes
+import hashlib
 import json
 import logging
 import os
@@ -323,11 +324,20 @@ def _download_with_cap(
     *,
     timeout: float,
     progress: Optional[ProgressCb],
-) -> bool:
-    """Stream the URL into ``dest`` with a hard byte cap."""
+) -> Optional[str]:
+    """Stream the URL into ``dest`` with a hard byte cap.
+
+    Returns the SHA-256 of the bytes actually written, or ``None`` if the
+    download failed.  The caller re-checks that digest once the file is
+    pinned, which is what closes the sliver between this function
+    releasing its write handle and the pin being taken; see
+    :func:`_pinned_for_execution` for the larger window either side of
+    it.  Returning the digest rather than a bool is the whole point of
+    the contract: only this function ever sees the bytes as they arrive.
+    """
     if not _is_safe_download_url(url):
         _logger.error("Refusing download from disallowed URL: %s", url)
-        return False
+        return None
 
     req = urllib.request.Request(url, headers={"User-Agent": _USER_AGENT})
     try:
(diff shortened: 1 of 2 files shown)


owenpkent:
Checked the blast radius of the `bool` -> `Optional[str]` return-type
change on `_download_with_cap`, since a changed contract is the thing
most likely to bite something this PR does not touch.

It is fully contained. One production caller:

- `src/updater.py:925`, inside `download_and_install` (updated here)

Every other reference in the repo is in `tests/test_updater.py`, all
updated in this PR. Nothing outside the module can reach it:
`_download_with_cap` is private, and the only import of `updater`
anywhere in `src/` is

```
src/keyboard_bridge.py:60:
from .updater import UpdateInfo, check_for_update, download_and_install
```

which takes the public surface only. `download_and_install`'s own
signature is unchanged (`Tuple[bool, str]`), so the bridge and the QML
update flow see no difference.

Verified with `rg -n "_download_with_cap"` across the tree, not by
reading the call site and assuming.

PR #84, opened 2026-09-06
feat(ui): one section height, and colour keys by what they do

owenpkent:
## Summary

Two halves of one change. The board was ragged vertically, and a key's colour
said which panel it sat in rather than what it does.

### Section heights

The grid, the nav cluster and the numpad sat at three different heights that
the `RowLayout` centred against one another, and the two separators filled
past both. Measured at a 1400 px window, that is five top edges and five
bottom edges:

| | top | bottom | height |
|---|---|---|---|
| separators | 105 | 433 | 328 |
| grid | 115 | 423 | 308 |
| nav | 130 | 408 | 278 |
| numpad | 134 | 404 | 270 |

Every section now lays out to `sectionHeight`, and the panels divide it
between their own five rows, so the **keys** grow rather than the gaps. Nav
and numpad targets come out about 12% taller, which is the reason that
option was picked over widening the gaps: those are the arrows and the
numpad. The arrow cluster lands on the bottom rail with nothing positioning
it there, because `arrowGap` already opened above the Up key.

The main grid is untouched in the default configuration, so `keyW`, the
window's width budget and every row's flush edges are exactly what they were.

### Key Colours

A picker under *Appearance* offering six schemes, **Monochrome by default**;
`off` restores the historical per-surface tinting in one click.

Every colour is derived from the active theme rather than written down:

- Family hues are rotated off the theme's own accent **in OKLCh, not HSL**.
  Rotating hue in HSL holds the number constant while perceived lightness
  swings, which produces bands where some shout and others whisper.
- Roles that carry meaning (Backspace/Del, Enter, the modifiers) keep an
  anchored hue pulled at most 22 degrees toward the accent, so a Delete key
  cannot come out green on Forest.
- Every fill walks its own strength down until the theme's text colour
  clears 4.5:1, the same rule `accentWashFor` already held the compact
  view's editing keys to.

Worst legend contrast across nine themes, five schemes and twelve roles:
**4.50:1**. Tightest hue pair anywhere: **30 degrees**.

Three defects were caught by checking the maths across all nine themes
before wiring any of it up, and each is worth knowing because they are the
traps a future change would walk back into:

- The Light theme collapsed to a single fill. Its key colour is `#ffffff`,
  and taking the hue's lightness from it leaves no room for chroma. Clamped.
- A tetradic harmony put `kill` 12 degrees from `edit` on Amethyst, i.e.
  Backspace and Tab the same colour on the one scheme whose purpose is
  telling them apart. A square has only one free phase angle and some themes
  have no phase that fits, so hues are now placed by farthest-point
  dispersion.
- The Ink scheme's punctuation legend sat at 4.32:1 on Vaporwave, under the
  bar the rest of the file promises, because a dim is not a fill and skipped
  the wash.

## Notes for review

- The colour engine is one new file, `qml/palette.js`, which is now also the
  **single copy of the WCAG contrast maths**; `Main.qml`'s
  `relativeLuminance` / `contrastRatio` / `accentWashFor` delegate to it.
- Role resolution lives in `KeyButton`, so the five key-drawing surfaces
  name a role instead of repeating the rule about fifty times.
- `off` hands down a **null** table that every surface reads as "keep your
  own tint". That is what makes the historical board reachable with no
  per-surface branch anywhere.
- Two decisions were made and then reversed on sight, and both are recorded
  in `CLAUDE.md` with what was wrong: giving the numpad a matching gutter so
  its rows would align with the nav cluster's (it reads as a seam splitting
  the digits off the `0` key), and washing Monochrome's modifiers with the
  accent (a modifier's theme colour is its *click* colour, not its resting
  one).
- No change to the prediction engine, the build/signing pipeline, or
  telemetry.

## Test plan

- `python check.py` passes: ruff, ruff-format, mypy under both `--platform
  linux` and `--platform win32`, and the full pytest suite.
- New `tests/test_qml_key_colors.py`, 47 tests against the real `Main.qml`
  loaded headlessly. The geometry half measures live item positions rather
  than recomputing them from the properties the QML sets; the colour half's
  contrast sweep covers every scheme x theme x role.
- Screenshots regenerated from the real UI via
  `scripts/capture_screenshots.py`.

## Accessibility

Nav and numpad keys are about 12% taller, so those targets get cheaper to
hit. Nothing changes about keystroke timing, repeat interval, or the
feedback on a press. Every scheme is held to 4.5:1 on every theme, and the
default scheme has no hue at all, so nothing here depends on colour
discrimination to be usable.

## Review fixes

A multi-angle review turned up the following, all fixed in the follow-up commit. The first item supersedes the *Section heights* description above.

- **The grid never took the shared height, so the shipped default was still ragged.** `sectionHeight` was a max over the grid's height and the panels' natural heights, meant to keep the panel keys at full height with the function row hidden. It could not work: the grid's own implicit height is what defines the section, so the grid cannot grow into a larger one, and it sat centred 6 px inside the panels in exactly the configuration a fresh install ships with (nav on, no function row, no numpad). Measured at 1400 px: grid 134-463 against nav 128-469. The section is now the grid's height and the panels fit it, with no floor on their row height. With a function row showing their keys are still about 12% taller than the letters; without one they come out about 2 px shorter. That is the trade for a straight edge, in the one configuration where the choice arises.
- **The test that claimed to cover that branch could not fail.** Qt Quick Layouts recompute in a polish step the offscreen window never runs on its own: after hiding the function row, `mainKeyboard.implicitHeight` sat at its old value through 300 event passes and moved only once a frame was forced. The test measured the board before its own toggle had taken effect. The geometry tests now force a frame with `grabWindow` after a toggle, assert the grid's span actually moved, and cover the shipped default configuration.
- **The hover lift broke the contrast promise.** A plain `Qt.lighter` on a fill the wash had left at exactly 4.5:1 overshoots the bar: Monochrome's Enter on Blackboard measured 3.1:1 under the pointer, the Ink scheme's pill on Light 2.9:1. `palette.js::hoverFill` walks the lift down until the legend still clears what it cleared at rest, and the pill's ink is no longer lifted with its fill. The sweep now covers hovered fills too.
- **The lightness step skipped the wash.** Monochrome and Function are built from `step()`, which was a bare blend; it goes through `washFor` like every other fill now. `legibleInk` picked its pole at 0.35 luminance rather than the 0.179 crossover, which for a ground between the two walked toward the pole that cannot reach 4.5:1 (no shipped theme sits there).
- **Compact's embedded nav keys were painted as editing keys.** The compact layouts carry Home, End, PgUp, PgDn, Insert and the arrows in the grid as `special` keys, which `roleForKey` read as `edit`, so Home was the same colour as Tab on the one layout that carries them this way. They are `nav` now, with a test on the compact grid.
- **Smaller:** the schemes are one builder each in a map whose keys are the known ids, so an unknown id reads as `off` rather than as an empty table; the family hues are built on first use (Monochrome, the default, never asks for them); the dead `relativeLuminance` / `contrastRatio` wrappers in `Main.qml` are gone; the numpad's NumLock role is one panel property rather than ten ternaries; `blockGap` and the duplicated natural-height formula are removed; the settings panel's default scheme matches the shipped one; a duplicated comment line is fixed; the parametrised tests that rebuilt the engine per case loop instead; CLAUDE.md's "297 combinations" was 540; changelog entries added.

**Left as is, for a look on screen:** in Compact View with a hue scheme, the accent-styled editing keys keep their full-accent border while their fill comes from the role table, so under Function a Backspace is a red-tinted key with a blue ring. Whether the ring should yield to the role fill is a visual call.


tests/test_qml_key_colors.py
```diff
@@ -0,0 +1,619 @@
...
+        before = {}
+        for key in _keys(root):
+            before[key.property("keyText")] = key.property("_roleFill").name()
+        root.setProperty("keyColorScheme", "bands")
+        _settle()
+        after = {}
+        for key in _keys(root):
+            after[key.property("keyText")] = key.property("_roleFill").name()
+        changed = [k for k in before if k in after and before[k] != after[k]]
+        assert len(changed) > 10, "Function bands repainted almost nothing"
```
owenpkent, line 474: This test, and every other colour test in the file, reads `_roleFill` /
`_roleInk` / `_roleBar`. Those are the *inputs* to the paint bindings, never the
painted result, so the suite cannot tell whether the table ever reaches the
screen.

Mutation: revert `KeyButton.qml:342-343` so the keycap paints `keyColor` instead
of `_roleFill`, and force `keyRoleBar.visible: false`. **All 47 tests still
pass.** The feature is entirely disconnected and the suite is green. Same shape
for the pills: `root.predPillFill` is asserted, the Rectangle at `Main.qml:2410`
that consumes it is not.

That is a shame, because the fix is small: `keyBackground` is reachable from the
existing `_descendants()` walk, and `keyRoleBar` already carries an
`objectName`. Asserting `keyBackground.color` on a resting key, and
`keyRoleBar.visible` for the `ink` scheme, would close it and would have made
this the strongest test in the file rather than the weakest.

tests/test_qml_key_colors.py
```diff
@@ -0,0 +1,619 @@
...
+        # the contrast sweep above perfectly.
+        root, _ = qml_root
+        themes = root.property("themeData").toVariant()
+        root.setProperty("keyColorScheme", "bands")
+        for theme in themes:
+            root.setProperty("currentTheme", theme)
+            _settle(3)
+            table = root.property("keyRoles").toVariant()
+            fills = {table[role]["fill"].name() for role in ROLES}
+            assert len(fills) >= 9, (
```
owenpkent, line 755: This bar is three collapses too loose to be the inverse it is offered as.

`bands` produces 12 distinct fills normally, so `>= 9` gives away three for
free. Mutation: force `_disperseHues` to return the same hue for all four family
roles (`edit`, `nav`, `fn`, `op`) and the suite still passes, on **all nine
themes**: the collapse only costs two distinct fills (10 remain), because `op`'s
0.28 wash strength keeps it apart from the other three at 0.30 even when the hue
is identical.

That is precisely the failure `_disperseHues` exists to prevent, and the one the
comment above it describes (the Amethyst case, `kill` 12 degrees from `edit`).
It is worth testing directly rather than through a fill count: assert on the
minimum pairwise `hueDistance` across the role hues, which also pins the 30
degree figure from the PR description. Nothing in the file currently references
`hueDistance`, `hueDelta` or `_disperseHues` at all.

A hex-`.name()` comparison is also a weak notion of "tellable apart", since one
step in 255 counts as distinct.

tests/test_qml_key_colors.py
```diff
@@ -0,0 +1,619 @@
...
+                for role in ROLES:
+                    face = table[role]
+                    ratio = _contrast(face["ink"], face["fill"])
+                    if ratio < worst[0]:
+                        worst = (ratio, f"{scheme}/{theme}/{role}")
+                    assert ratio >= 4.49, (
+                        f"{scheme} on {theme}: the {role} legend is only "
+                        f"{ratio:.2f}:1 on its own keycap"
+                    )
+        assert worst[0] >= 4.49, worst
```
owenpkent, line 716: This assertion cannot fail. `worst` is only ever assigned a `ratio` two lines
above the in-loop `assert ratio >= 4.49`, so any value low enough to trip it has
already failed there. And if the sweep body never ran, `worst` is still
`(99.0, "")` and it passes vacuously.

Harmless, but it reads as a second, stricter gate and is neither. If the intent
is to record the margin, printing `worst` or asserting a headroom figure would
do something; as written it is decoration.

qml/palette.js
```diff
@@ -0,0 +1,464 @@
...
+    if (kd.type === "special") {
+        switch (kd.action) {
+        case "backspace":
+        case "delete":
+            return "kill"
+        case "return":
+        case "enter":
+            return "commit"
+        default:
+            return "edit"
```
owenpkent, line 506: The space bar lands here.

`{type: "special", action: "space"}` matches no case, so it falls through to
`default:` and comes out `edit`. On the shipped default (Mono on Dark) that
moves the space bar from `#3a3a3a`, which is pixel-identical to the letters
today, to `#292929`, the same fill as Tab, Esc, Shift and Backspace. Mono,
Two-Tone and Function all repaint it; Signal leaves it flat.

It is the most-pressed key on the board and it does type a character, so I do
not think it should pick up "does something" colouring by falling through a
default arm. Either give it its own case returning `alpha`, or add a case
returning `edit` explicitly so the decision is recorded. As it stands it reads
as unconsidered rather than chosen.

qml/palette.js
```diff
@@ -0,0 +1,464 @@
...
+        map[role] = {
+            fill: fill,
+            ink: ink === undefined ? text : ink,
+            bar: bar === undefined ? clear : bar
+        }
+    }
+    // A lightness step toward the page behind the keys. Direction-agnostic:
+    // on a light theme it lightens, on a dark one it darkens, and either
+    // way it reads as one step away from the alpha field.
+    function step(t) { return mixToward(key, background, t) }
```
owenpkent: `step()` is the one fill path that never goes through `washFor`, so it is not
contrast-guarded.

It feeds `mono`'s punct/nav/op (0.30) and mod/edit/kill/fn (0.52), and the base
of `twotone`'s `util`. The PR description and the new CLAUDE.md bullet both say
every fill walks its own strength down until the theme's `textColor` clears
4.5:1, and that is not true of these.

It holds on all nine shipped themes, so nothing is wrong today. It is reachable
though: a theme with `#fafafa` keys on a `#9a9a9a` page with `#5a5a5a` ink puts
the alpha keys at a comfortable 6.61:1 and drives `step(0.52)` to **4.13:1**,
under the bar, with nothing about the theme looking marginal.

The sweep's `assert len(themes) == 9, "theme count changed; this sweep should
follow it"` is what actually protects this, since a tenth theme fails the suite
and forces someone back through the sweep. So I would call this a docs accuracy
issue rather than a live risk. Either route `step` through `washFor` as well, or
narrow the claim in both places to say which fills are guarded.

qml/Main.qml
```diff
@@ -1051,30 +1101,75 @@ Window {
...
+    // Read that file before changing a scheme.  Two things there are
+    // load-bearing and neither is obvious from here: every family hue is
+    // rotated off the ACTIVE THEME'S accent rather than being a literal
+    // (nine themes ship, and a fixed hex is unreadable on about half of
+    // them), and the rotation happens in OKLCh rather than HSL, because
+    // rotating hue in HSL holds the number constant while perceived
+    // lightness swings, which produces bands where some shout and others
+    // whisper.
+    //
+    // `keyRoles` is null for the default "off" scheme, and every surface
```
owenpkent, line 1130: This says the default scheme hands down null, and that it is what keeps the
shipped board byte-identical to the one before this existed. The default is
`mono` though (line 47, and `test_the_default_scheme_is_monochrome` pins it).
`off` is what hands down null, and `off` is not the default.

The sentence is correct about `off`; it is only "the default" that is wrong.
The same wording reached the CLAUDE.md bullet, flagged separately.

CLAUDE.md
```diff
@@ -19,6 +19,8 @@ Owen is a wheelchair user with muscular dystrophy. Typing is hard - be proactive
...
 - Telemetry is OFF by default and `DEFAULT_ENDPOINT` in `src/telemetry.py` ships empty (silent no-op). `TelemetryClient` is the source of truth for the consent flag; do NOT mirror it into `appSettings`. The Data Backup archive deliberately excludes `telemetry.json`.
+- Key colouring is a role -> colour table (`qml/palette.js`) that `Main.qml` hands to every keyboard surface, and `KeyButton` resolves; the default scheme hands down `null`, which every surface reads as "keep your own tint" (Monochrome is the shipped default). No hue is ever a literal: family hues are rotated off the active theme's accent in **OKLCh** (HSL swings perceived lightness), and every fill goes through the same wash that walks its strength down until the theme's `textColor` clears 4.5:1. See *Key Colours by role*.
```
owenpkent, line 23: This contradicts itself inside one sentence: "the default scheme hands down
`null`", then "(Monochrome is the shipped default)". Monochrome does not hand
down null. `off` does.

It matters more here than in the source comment, because this is *Key rules*,
which is what a future reader takes as authoritative. As written it says the
change is inert by default, when in fact it repaints every existing user's
keyboard.

The prose further down, in the *Key Colours by role* section, already states
this correctly, so it is just this bullet.

qml/Main.qml
```diff
@@ -772,6 +780,48 @@ Window {
...
+    // not slop: a `Row` reports a height ceiled above its tallest key (see
+    // `keyHitMarginV`), so a panel's natural height is the ceiled one and
+    // computing it any other way leaves the panels a pixel short.
+    readonly property real panelBlockGap: keySpacing * 4
+    readonly property real _panelNaturalH: 5 * Math.ceil(keyH) + 4 * keySpacing
+                                           + panelBlockGap
+    readonly property real sectionHeight: Math.max(
+        mainKeyboard.implicitHeight,
+        showNavigation ? _panelNaturalH : 0,
+        showNumpad ? _panelNaturalH : 0)
```
owenpkent: `_panelNaturalH` adds `panelBlockGap` for both panels, but the numpad does not
have one: `NumpadPanel.blockGap` defaults to 0 and nothing here sets it, so the
numpad's real natural height is `5 * Math.ceil(keyH) + 4 * keySpacing`, about
8 px less at the default width.

With Navigation off, Numpad on, and the grid shorter than that value,
`sectionHeight` overshoots what any section actually needs, and the grid then
gets centred inside it. That is the ragged edge this PR exists to remove,
reappearing in that one configuration.

`showNumpad ? _panelNaturalH - panelBlockGap : 0` would cover it, or give the
numpad its own natural-height expression so the two panels stop sharing one.

qml/components/KeyButton.qml
```diff
@@ -148,6 +148,38 @@ Item {
     // transient UI such as the key-preview bubble.
     signal keyReleased()
 
+    // ===== Colour by role =====
+    //
+    // `role` is what this key *does*: "alpha", "digit", "punct",
+    // "mod", "edit", "kill", "commit", "nav", "fn", "op", "toggle".
+    // `roleColors` is
+    // `roleColors` is the {role: {fill, ink, bar}} table Main.qml derives
```
owenpkent, line 156: Line 155 is a truncated duplicate of this line: "`roleColors` is", then
"`roleColors` is the {role: {fill, ink, bar}} table ...". Looks like a paste
that lost its tail.

qml/components/UnifiedSettingsPanel.qml
```diff
@@ -14,6 +15,25 @@ Item {
     property bool showNumpad: false
     property string currentTheme: "dark"
     property var themeData: ({})
+    // Which Key Colours scheme paints the keycaps; see qml/palette.js.
+    property string keyColorScheme: "off"
```
owenpkent: Defaults to `off` here, while `Main.qml`'s `savedKeyColorScheme` defaults to
`mono`.

Harmless today, since `Main.qml` always binds the real value over it. But the
8-step wiring convention in CLAUDE.md has the panel's own fallback match the
saved default, and this fallback is exactly what would show if that binding
were ever dropped.

owenpkent (commented):
Read the whole diff, and ran mutation tests against the new suite rather than
only reading it.

**The colour engine itself checks out.** To verify the two headline numbers
independently I re-derived `palette.js` in Python against the real nine themes,
rather than trusting the sweep in the test file. Both reproduce exactly: worst
legend contrast **4.503:1** (dark / mono / toggle), tightest hue pair **30.0
degrees** (spaceship, fn/op). The OKLCh conversion, the gamut walk that reduces
chroma instead of clamping channels, and the farthest-point dispersion are all
correct, and the reasoning recorded for why dispersion replaced the tetrad is
borne out.

Two habits in the test file are right, and worth naming because both are traps
this repo has paid for before: `_keys()` walks `childItems()` recursively
instead of `findChildren`, so it really does iterate the Repeater delegates; and
`_contrast()` reimplements the luminance maths in Python rather than calling
into `palette.js`, so the contrast sweep is independent verification rather than
arithmetic agreeing with itself. That sweep is genuinely load-bearing: removing
`washFor`'s contrast loop drops it to 3.97:1 and it fails.

**But two mutations pass the whole suite**, and both are on the comments below:
disconnecting the paint bindings entirely, and collapsing every family hue onto
one. Those are the two things this PR is for, and neither is currently pinned.

The rest of the findings are text, or a narrow geometry edge case, or a
judgement call about the space bar.

One thing that is not a line comment: this repaints every existing user's board
on upgrade, because the default is `mono` rather than `off`. That is defensible
and the PR argues for it, but it deserves a CHANGELOG line so it is announced
rather than discovered.

Smaller coverage gaps, none worth blocking on: `roleForKey`'s
`type === "layer"` branch and the new `NumberRow.qml` role wiring have no
coverage, because no compact layout is loaded anywhere in the module; the
settings picker (+163 lines) is untested, including that its scheme ids match
`palette.SCHEMES`; and `test_monochrome_really_is_monochrome` runs on the
default theme only, while its `neutral` baseline is the theme's own key-colour
saturation, so the `+ 0.02` bar is loosest on exactly the saturated themes
(Ocean, Vaporwave) where `mono`'s unguarded `step()` mixes toward `background`
could let a hue in.

PR #85, opened 2026-09-07
fix(ui): stop the white notch in the window corner on Windows

owenpkent:
## Summary

A small bright notch bites into a corner of the keyboard on Windows. It
comes and goes depending on what is behind the window, which is why it reads
as intermittent and unrelated to anything.

Every window in this app is frameless and `WS_EX_LAYERED`, and each rounded
its own corners with a `radius` on the QML background rectangle. That leaves
the pixels outside the arc unpainted, and **on a layered window those do not
composite the desktop the way a transparent pixel should: they come back
white**.

## Evidence

Measured on the real `Main.qml`, with a magenta backdrop placed behind all
four corners so "what shows through" is a colour nothing else on the desktop
would be:

| configuration | corner pixel |
|---|---|
| radius 10, Windows 11 default rounding (**ships today**) | `#ffffff` |
| radius 10, `DWMWCP_DONOTROUND` | `#ffffff` |
| radius 0, `DWMWCP_ROUND` | the backdrop, correctly |

The middle row is the useful one: **turning Windows' own rounding off fixes
nothing.** The notch is the unpainted region, not the rounding, so the fix
has to be to leave nothing unpainted.

After the change, all four corners of the real window report **zero** white
pixels and show the backdrop, DWM-rounded and antialiased.

## What changed

- `Main.qml` gains `selfRoundedCorners` (false on Windows) and
  `windowRadius`, applied to the background, the title bar and the shadow.
  The title bar has to follow the background: rounding it over a square
  background swaps the notch for a lighter wedge, since what shows through
  is then the background rather than the desktop.
- `windows_window.py::_prefer_dwm_rounded_corners` sets
  `DWMWA_WINDOW_CORNER_PREFERENCE` to `DWMWCP_ROUND`, so the compositor
  masks an opaque window.
- `SnippetsWindow` and `SymbolsWindow` are transparent with a radius-8
  background in exactly the same way, so they take the same decision as a
  required property from `Main.qml` rather than each reading
  `Qt.platform.os`. A fix reaching only the keyboard would have left the
  notch on the two windows that float over whatever the user is typing into.

## Notes for review

- **The DWM call is best-effort and should stay that way.**
  `DWMWA_WINDOW_CORNER_PREFERENCE` is Windows 11 and later; on Windows 10 it
  fails and the window keeps the square corners QML gave it, which is what
  every other window on that desktop looks like. A window that fails to
  round is cosmetic, never a reason to fail startup. It carries the same
  `sys.platform` guard every other function in that file has, so mypy prunes
  the `ctypes.windll` body under `--platform linux`.
- Nothing changes off Windows: `selfRoundedCorners` is true there and the
  radius stays 10.

## Test plan

- `python check.py` passes: ruff, ruff-format, mypy under both platforms,
  full pytest.
- New `tests/test_window_corners.py`. **It deliberately does not render the
  corner and look at it**: the offscreen platform plugin has no compositor,
  and the offscreen render was correct the whole time this was visible on
  screen, so such a test would have passed against the bug. It pins what is
  checkable instead: which side is asked to round, that the rectangle is
  actually bound to the property rather than carrying a stale literal, that
  all three windows agree, and that the DWM call cannot raise.
- Verified by hand on the real window, over a known backdrop, before and
  after.

## Review fixes

A multi-angle review of the first commit turned up the following, all fixed in the follow-up commits:

- **The Dashboard window was missed.** It is transparent with a radius-10 panel in exactly the way the two pickers are, so the notch was still there on *Your Language Model*. It now binds `selfRoundedCorners` like them and is named in `_wire_floating_windows`, where it gets only the DWM corner call: it is allowed to take focus (it has no keys on it), so it must not go through `apply_extended_styles` and pick up `WS_EX_NOACTIVATE` with the corner.
- **The DWM call sat after two early returns** in `apply_extended_styles`, so a failed style read or write would have cost the corner too. It runs before them now, and a test drives that path.
- **Screenshots would have regenerated square.** `Qt.platform.os` still reads `"windows"` under the offscreen plugin, where there is no compositor to round anything, so `capture_screenshots.py` sets `selfRoundedCorners` back to true after loading `Main.qml`. The property is no longer `readonly` for that one caller.
- **The CLAUDE.md claim was too broad.** "Every window is layered" is false for Settings and Help, which are opaque and whose rounded panels show the window's own dark colour in the corners rather than white; the rule is narrowed to the four transparent windows and says what a new one has to do.
- **The measurement table was restated in five places.** It now lives once, in CLAUDE.md, with the code comments pointing at it.
- **Tests.** Loaded once per module, assert on QML warnings like their siblings, use `findChild(QObject, name)`, check what the DWM call asks for (attribute 33, value 2) and that the dashboard path never touches the style word, and run on every platform through a fake `windll` (which needed `create=True` on the Linux shards). The existing window-style harness gains a `dwmapi` so it stops exercising the failure path unannounced.

**Not verified on a real desktop:** the background keeps its 1 px theme border while its radius is 0, so along the corner arc DWM's mask clips that border and draws its own. If that reads wrong on a light theme, the answer is `DWMWA_BORDER_COLOR`, not a radius on the QML side. Noted in CLAUDE.md.

### DIFF
--- a/tests/test_window_corners.py
+++ b/tests/test_window_corners.py
@@ -1,0 +1,170 @@
+"""Who rounds the window's corners.
+
+Every window in this app is frameless and, on Windows, ``WS_EX_LAYERED``.
+Rounding a corner ourselves with a QML ``radius`` leaves the pixels outside
+the arc unpainted, and on a layered window those do not composite the
+desktop the way a transparent pixel should: they come back white. What the
+user sees is a bright notch biting into a corner of the keyboard.
+
+Measured on the real ``Main.qml`` against a magenta backdrop placed behind
+all four corners:
+
+===========================================  =========================
+configuration                                corner pixel
+===========================================  =========================
+radius 10, Windows 11 default rounding       ``#ffffff``
+radius 10, ``DWMWCP_DONOTROUND``             ``#ffffff``
+radius 0, ``DWMWCP_ROUND``                   the backdrop, correctly
+===========================================  =========================
+
+Turning Windows' own rounding off fixes nothing, which is the part worth
+remembering: the notch is the *unpainted region*, not the rounding. So on
+Windows the background is squared off and DWM masks the corner instead.
+
+**The compositing itself cannot be tested here.** The offscreen platform
+plugin has no compositor, and the offscreen render was correct the whole
+time this bug was visible on screen, so a test that rendered the corner and
+looked at it would have passed against the bug. What these tests pin is the
+half that is checkable: which side is asked to do the rounding, that every
+window agrees about it, and that the DWM call can never break startup.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+pytest.importorskip("PySide6")
+
+os.environ.setdefault("QT_QPA_PLATFORM", "offscreen")
+
+try:
+    from PySide6.QtCore import QCoreApplication, QSettings, QUrl  # noqa: E402
+    from PySide6.QtGui import QGuiApplication  # noqa: E402
+    from PySide6.QtQml import QQmlApplicationEngine  # noqa: E402
+    from PySide6.QtQuick import QQuickItem  # noqa: E402
+except ImportError as exc:  # pragma: no cover - environment-dependent
+    pytest.skip(
+        f"Qt GUI libraries unavailable ({exc}); install libegl1/libgl1 to run "
+        "the headless QML tests",
+        allow_module_level=True,
+    )
+
+from src.keyboard_bridge import KeyboardBridge  # noqa: E402
+from tests.qml_context import install_context_properties  # noqa: E402
+from tests.qt_settings_scope import TEST_APP, TEST_ORG  # noqa: E402
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+QML_MAIN = REPO_ROOT / "qml" / "Main.qml"
+
+ON_WINDOWS = sys.platform == "win32"
+
+
+@pytest.fixture(scope="module")
+def qapp():
+    app = QGuiApplication.instance()
+    if app is None:
+        QCoreApplication.setOrganizationName(TEST_ORG)
+        QCoreApplication.setApplicationName(TEST_APP)
+        app = QGuiApplication([])
+    assert QCoreApplication.organizationName() == TEST_ORG, (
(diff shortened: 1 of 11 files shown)


tests/test_window_corners.py
```diff
@@ -0,0 +1,170 @@
...
+        assert background.property("radius") == root.property("windowRadius")
+
+    def test_every_window_agrees(self, qml_root) -> None:
+        # The snippets and symbols windows are frameless and transparent in
+        # exactly the same way, so a fix that reached only the keyboard
+        # would leave the notch on the two windows that float over the app
+        # the user is typing into.
+        root = qml_root
+        expected = root.property("selfRoundedCorners")
+        for name in ("snippetsWindow", "symbolsWindow"):
```
owenpkent: There is a third window with exactly this shape, and it is not on this list.

`vizWindow` (`Main.qml` ~4904, the Model Visualization / dashboard window) is
frameless and `color: "transparent"`, and `ModelVisualization.qml:71-74` fills
it with a `radius: 10` rectangle. That is the same combination this PR
identifies as the cause: translucent, so Qt makes it layered, plus a radius
leaving the corner pixels unpainted. So the notch should still be on the
dashboard after this lands.

Worth knowing before fixing it: `vizWindow` is not wired through
`_wire_floating_windows` either, since that loop names only `snippetsWindow`
and `symbolsWindow`. So squaring its background off would leave it with square
corners rather than DWM-rounded ones. That is still strictly better than a
white notch, and it is the same trade this PR already accepts on Windows 10.

`settingsWindow` and `helpWindow` are fine as they are: both opaque `#1e1e1e`.

owenpkent (commented):
Reviewed the whole diff, plus the three windows it does not touch.

The diagnosis holds up, and the evidence table is the part that makes it
convincing: ruling out `DWMWCP_DONOTROUND` before landing the real fix is what
shows the notch is the unpainted region rather than the rounding. The test
docstring is also honest about what the offscreen plugin cannot check, which
matters here because the offscreen render was correct the whole time the bug
was on screen.

One gap, in the comment below. The rest reads correctly to me: the two floating
windows do reach `_prefer_dwm_rounded_corners`, through `_wire_floating_windows`
-> `apply_extended_styles`; the title bar following the background is right for
the reason given; and the DWM call carries the same `sys.platform` guard and
best-effort handling as everything else in that file.

For the record, `settingsWindow` and `helpWindow` need nothing: both are
`color: "#1e1e1e"`, so Qt never makes them layered and they have no unpainted
region to begin with.

PR #95, opened 2026-09-07
fix(ci): publish the telemetry series to a branch main's rules allow

owenpkent:
The aggregate workflow could never write a row. It had **two independent faults**, and the first hid the second.

## Fault 1 (fixed in #94)

The commit step asked `git diff --quiet` about a file that is **untracked** on the first run. The diff came back empty, so the step printed `No change.` and exited before ever reaching `git add`.

That run **reported success and committed nothing**, which is why it survived review: nothing in the log said anything was wrong.

## Fault 2 (this PR)

Fixing the first exposed the real blocker. The push is rejected outright:

```
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: - 4 of 4 required status checks are expected.
 ! [remote rejected] main -> main (protected branch hook declined)
```

main requires four status checks, and a bot appending one CSV line cannot satisfy them.

## The fix, and why this shape

The series now lives on the **orphan branch `telemetry-data`**.

The alternatives were both worse:

- **A pull request per day** would run the full eight-job suite against a two-line data change, 365 times a year, and make the series hostage to a flaky shard.
- **A protection bypass for the bot** would weaken main's guarantees for the sake of a file that is machine-generated data rather than code.

**Orphan** rather than a branch off main: it carries none of main's tree, holds only the series, cannot be mistaken for a feature branch, and cannot be merged into main by accident.

Read the series at:

```
https://raw.githubusercontent.com/owenpkent/alpha-osk/telemetry-data/docs/research/data/telemetry-aggregate.csv
```

## Three details worth keeping

- **The endpoint is read before any git work**, so a bad response fails the run without leaving a half-switched checkout behind.
- **The checkout is full-depth**, because a shallow clone can have its push rejected as a shallow update, which would have been a third variant of this same bug.
- **The reasoning is written into the workflow header**, because the obvious "fix" is to move this back onto main, which is exactly where it started.

## Why this matters now

Telemetry went live in #94, so the endpoint is collecting. But `/v1/aggregate` returns **current totals and keeps no history**, so until this works there is no series at all, and any period it misses cannot be reconstructed. There are 0 opted-in users today, so nothing has been lost yet, and that is the whole reason to fix it before 1.4.0 ships rather than after.

## Verification

Both faults were invisible from the source and only showed on a real dispatch. I will dispatch again after merge and confirm the branch and the first row exist, then report back. YAML parse checked locally.

## Accessibility

No UI change.


owenpkent:
Superseded by #97, which contains the workflow change on its own.

This branch accidentally picked up the unrelated two-letter suggestions commit, so the PR held two independent changes. The branch is deliberately left in place rather than force-pushed, so nothing is removed from the remote.

PR #103, opened 2026-09-08
fix(installer): the shortcut boxes forgot what the user chose

owenpkent:
Found while reviewing #102. Same defect as the consent checkbox on the page after this one, and the same cause.

## The bug

`ShortcutOptionsPage` set both checkboxes from a literal `BST_CHECKED` on every entry. NSIS rebuilds a custom page's dialog each time the page is shown, so:

1. Untick "Create Desktop shortcut"
2. Click Back
3. Click Next

The box is ticked again, `ShortcutOptionsLeave` reads the box rather than the answer, and the shortcut is created against the user's wishes. Nothing on screen says so.

## The fix

Seed each box from its variable instead. `.onInit` already sets both to `BST_CHECKED`, so the variable is the default on a first visit and the user's own answer on any later one. That is why this needs no guard, unlike the consent box, which has to distinguish "not answered yet" from "declined".

## Testing

Three tests, and the second is the one that carries its weight: seeding a box from a variable that nothing defaults would leave both shortcuts **off** on the only visit most users make, while satisfying the positive assertion perfectly.

Mutation-checked, each failing only its intended test:

| Mutation | Fails |
|---|---|
| back to the literal `BST_CHECKED` (the original bug) | `test_each_box_is_seeded_from_its_variable` |
| drop `.onInit`'s default for the desktop box | `test_the_variables_still_default_to_checked` |

The generated `.nsi` also compiles clean under makensis 3.12 with no warnings, which is the half the source tests cannot check.

## Notes for review

- No behaviour change on a first visit: both boxes still arrive ticked.
- `$1` / `$2` are still the scratch registers the page always used for the handles. Switching those to dedicated `Var`s, as the invite page does with `$StudyInviteCheckboxHwnd`, would be tidier but is a separate change.


owenpkent:
Added a second commit while verifying the first.

**The label's handle was never popped.** Every `${NSD_Create*}` is a `nsDialogs::CreateControl` call and pushes the new control's HWND, so one without a `Pop` strands an item on the stack for the life of the process, once per visit to a page the user can reach twice. `ShortcutOptionsPage` created its label and never popped it, and it was the only site in the file that did: the invite page pops all five of its controls.

The two checkboxes were unaffected, because the stack is LIFO and their own `Pop`s took the handles they had just pushed. So this is untidiness rather than a live failure, but it is one line and this PR is already open on that function.

The test sweeps **every** custom page function rather than pinning this one line, since the point is to stop the next page being written the same way. It asserts it found some pages before sweeping them, because a regex that matched nothing would otherwise report a clean sweep over an empty set.

Mutation-checked: removing the `Pop` again fails the sweep and names the exact site.

I looked at the `$1` / `$2` scratch registers holding the checkbox handles across `Show` and deliberately left them. Nothing runs between `nsDialogs::Show` and the leave function on this page (no `NSD_OnClick` handlers), so the handles are intact; it is a latent footgun for whoever adds a click handler later, not a defect today, and swapping in dedicated `Var`s the way the invite page does with `$StudyInviteCheckboxHwnd` belongs in its own change.

PR #107, opened 2026-09-10
fix(installer): preserve settings through legacy upgrades

owenpkent:
## Summary

- Preserve preferences when upgrading from releases through 1.3.0, whose uninstallers delete the settings registry tree even in silent mode. Copy the full tree before cleanup and restore it before setup continues, including when the old process fails.
- Stop before cleanup if backup fails; retain a recovery copy and stop if restoration fails. Apply the guard to both same-directory and different-directory upgrades.
- Add executable NSIS regression tests, update upgrade documentation, create AGENTS.md from CLAUDE.md, and ignore local test caches and generated installer checks.
- Update the telemetry worker's development dependency overrides and lockfile to patched `js-yaml` 4.3.2 and `sharp` 0.35.4, resolving the dependency scan findings.
- Add named test results and a five-minute test-worker watchdog to CI so a stalled Windows test yields thread stacks and a failure instead of an anonymous job timeout.

## Related issue

No linked issue.

## Type of change

- [x] Bug fix
- [x] Build / CI / tooling

## Test plan

- [x] GitHub CI passes on the final commit: all eight Windows/Linux test shards, both coverage gates, lint, type checks, OSV, and Socket Security.
- [x] `python check.py` passes locally: lint, formatting, type checks for Linux and Windows, and the full test suite.
- [x] Added or updated tests.
- [x] Verified on Windows using compiled NSIS harnesses with isolated registry keys: typed nested values, failed uninstall, absent settings, backup failure, and restoration failure.
- [x] Generated installer compiler smoke check passes with application payload omitted.
- [x] Worker clean install, TypeScript check, and full npm audit pass (zero vulnerabilities, including development dependencies).
- [x] Windows shard 1 passes locally with CI coverage settings: 639 passed, one skipped. Sharding tests and a deliberate stalled-worker probe verify the new watchdog.

App launch and Linux/macOS runtime checks are not applicable to this Windows installer change.

## Accessibility check

- [x] No new requirement for fast or precise input.

No changes to keyboard layout, input timing, or visual feedback.

## Notes for reviewers

The guard copies registry values directly rather than importing a registry file into the elevated installer. Old uninstall commands include `_?=` so restoration waits for the actual uninstaller to finish. The fix preserves settings present at upgrade time; it cannot reconstruct preferences deleted by an earlier update.


owenpkent:
## Review

The settings-preservation logic itself is sound, and the `RegCopyTreeW` backup/restore pair is the right shape. The problem is `_?=`, which changes three behaviours that the surrounding code was relying on being asynchronous.

### 1. The app disappears from Add/Remove Programs after a different-directory upgrade

`_?=` plus `ExecWait` makes the previous uninstaller genuinely synchronous. It was not before: the stub returned as soon as it spawned its `%TEMP%` copy, so the ordering below was a race the stub usually lost.

Ordering in the generated script (`build/windows/build.py`):

- 975: `WriteUninstaller "$INSTDIR\uninstall.exe"`
- 976-987: the `WriteRegStr HKCU "...\Uninstall\${APP_GUID}"` block for Add/Remove Programs
- 1009: `!insertmacro customInstall`, which is where `RunPreviousUninstaller` now blocks

`APP_GUID` is the fixed literal `alpha-osk-keyboard` (build.py:612), identical in every release, and every uninstall section ends with `DeleteRegKey HKCU "...\Uninstall\${APP_GUID}"` (build.py:1054). So the old uninstaller is now guaranteed to finish inside `customInstall` and delete the ARP entry the new install wrote thirty lines earlier.

Result: after a different-directory upgrade Alpha-OSK is absent from Add/Remove Programs, and the next installer's own HKCU detection finds nothing to clean up.

Fix is either to move `customInstall` above the ARP writes, or to rewrite the ARP block after it.

### 2. The old install directory is left behind permanently

NSIS documents that `_?=` makes the uninstaller run in place rather than copying itself to `%TEMP%`, and that it therefore cannot delete itself and does not remove the install directory.

The same-directory branch already handles this (build.py:967, `Delete "$INSTDIR\uninstall.exe"` with a comment saying exactly why). Neither different-directory branch in `installer.nsh` does:

- `removePrevHKCU` (installer.nsh:189-192): `Push '"$0" /S _?=$2'`, `Call RunPreviousUninstaller`, `Sleep 2000`, and nothing else
- `removePrevHKLM` (installer.nsh:207-210): same

Before this change the temp-copy path removed both the uninstaller and the folder. Now `$2\uninstall.exe` and its directory survive forever. Both branches want a `Delete "$2\uninstall.exe"` and an `RMDir "$2"` after the call.

### 3. The new `Abort` lands mid-install and leaves no shortcuts

`RunPreviousUninstaller` aborts on a non-zero exit or a launch failure (upgrade_settings.nsh:89). Called from `customInstall` at build.py:1009, that abort sits between:

- above it: `{files_block}`, `WriteUninstaller`, the ARP writes, the study-invite seed
- below it: every `CreateShortCut` (build.py:1017, 1018, 1022)

So the failure state is a fully extracted and registered install with no Start Menu entry, no desktop icon, and (per finding 1) no ARP entry either. For a user whose only input device is this keyboard, the app is then reachable only by typing its full path.

The trigger does not need to be exotic: `IfFileExists` passes but the exe cannot launch (AV quarantine, SmartScreen, a truncated file), `ExecWait` sets the error flag, line 61 forces `$6 = -1`, line 89 aborts. On the same-directory path this runs before `{files_block}`, so every retry fails identically at the same point, and the auto-updater drives exactly this path with `/S`. Before this PR the return code was discarded and the install proceeded.

Worth considering whether an uninstaller failure should abort at all, given what it now costs. Logging and continuing is closer to the previous behaviour and to how the rescue-export failure is handled in `data_export.py`.

### 4. Under credential elevation the fix is a silent no-op

The installer runs elevated (`run.py::ensure_admin_windows`, and `INSTALL_DIR` is `$PROGRAMFILES64`). When a standard user elevates by typing an administrator's credentials, the elevated process's `HKEY_CURRENT_USER` resolves to the *admin's* profile, not theirs.

So `RegOpenKeyExW(HKEY_CURRENT_USER, "Software\alpha-osk")` at upgrade_settings.nsh:23 reads a tree that is usually absent, takes the `$1 == 2` shortcut, and the whole protection does nothing. Where an admin-profile key does happen to exist, the restore at line 68 writes `Software\alpha-osk` into a profile that is not the keyboard's user.

CLAUDE.md already records this hazard as the reason the study-invite seed goes to HKLM: "it runs elevated, possibly under a different account than the person who will use the keyboard". The new text in `docs/build/WINDOWS.md` and `CHANGELOG.md` states the preservation works without qualifying it.

### 5. Failure messages are suppressed on the one path that matters

On a silent auto-update, `/SD IDOK` auto-answers the MessageBox so it never appears, `DetailPrint` (line 78) goes to a details pane silent mode never shows, and `src/updater.py` only sees an exit code. If the restore fails, the recovery copy sits at `HKCU\Software\alpha-osk-upgrade-backups\nsXXXX.tmp` under a random temp name the user has no way to learn, and the keyboard comes back with default theme, layout and panels with no explanation.

`alpha-osk.log` is the file CLAUDE.md designates for exactly this and nothing on this path writes to it. Worth at least writing the retained key path somewhere recoverable.

### 6. The tests cannot fail if `_?=` is removed

`_new_source` (tests/test_upgrade_settings.py:78) pushes `'"{old}" /S'` with no `_?=`, and `_old_source` builds a `SilentInstall silent` *installer* rather than an NSIS uninstaller, so the stand-in never performs the copy-to-`%TEMP%`-and-return-immediately behaviour that `_?=` exists to suppress.

Deleting `_?=$INSTDIR` from build.py:964 and `_?=$2` from installer.nsh:190/208 leaves all four tests in this file passing. Only the string matching in `test_windows_installer.py` catches it, and that cannot observe the race itself.

Per the project's own doctrine on `TestTheClearButtonIcon` and the prediction-bar tests, a check that cannot fail is not a check. Worth building the stand-in as a real uninstaller so the suite can exercise both orderings.

Related gap: no test covers "no backup taken" plus "uninstaller exits non-zero", which is the cell this PR changes most. `test_legacy_uninstaller_preserves_typed_nested_settings` parametrises exit codes 0 and 9 but always seeds the settings key first, so the abort at line 89 is only ever reached after a successful restore. In the uncovered cell the abort fires while the MessageBox claims "Your saved settings have been preserved", about settings that never existed.

### Smaller items

- **`AGENTS.md` is a 2,492-line byte-identical copy of `CLAUDE.md`** (only the H1 differs) with nothing keeping them in sync. This PR already has to make the same settings-key edit in both. CLAUDE.md names parallel copies of one fact as the project's recurring failure mode and prescribes one source with every reader pointed at it. A one-line `AGENTS.md` pointing at `CLAUDE.md`, or a test asserting the two match, costs nothing.
- **`upgrade_settings.nsh` has no include guard.** It defines a `Function` at top level, so a second `!include` is a compile error rather than a no-op. Both headers it includes itself (LogicLib, FileFunc) carry `!ifndef` guards for this reason. Relevant to the NSIS harness pattern CLAUDE.md prescribes, which lifts functions out of the generated script.
- **Two of the three backup-failure paths leak the parent key.** `RegCreateKeyExW` at line 34 creates `Software\alpha-osk-upgrade-backups` as a side effect. The copy-failure path (51-53) removes the leaf and the empty parent; the disposition path (40-42) and the create-failure path (36-37) `Goto backupFailed` without either.
- **A retained recovery copy can brick future installs.** The key name comes from `${GetFileName} "$PLUGINSDIR"`, and `GetTempFileName` is only unique against files currently in `%TEMP%`. After a temp clean, a later run can be handed the same name, `RegCreateKeyExW` returns `REG_OPENED_EXISTING_KEY`, and line 39 aborts with "Please retry setup", which will fail identically on every retry until the key is deleted by hand. Nothing ever consults or cleans a retained copy.
- **The `Sleep` calls are now dead weight.** `Sleep 2000` at installer.nsh:192 and :210 plus `Sleep 1500` at build.py:966 only ever covered the window between the stub returning and the temp copy finishing. `ExecWait` now blocks until the real uninstaller exits, so they add 3.5s to 5.5s of wall clock on the upgrade path and suggest to the next reader that the wait is still load-bearing.
- **`faulthandler_timeout` is armed only inside `pytest_runtest_protocol`** (`_pytest/faulthandler.py` arms in a wrapper and cancels in its `finally`), so it is disarmed during collection, session-scoped fixture teardown, pytest-cov's end-of-session write, and xdist shutdown. The comment's own symptom, "a Windows shard can stall after nearly all tests finish", points at exactly that window, so the change may only have bought the `-vv` test names.

build/windows/upgrade_settings.nsh
```diff
@@ -0,0 +1,104 @@
...
+      SetErrorLevel 1
+      Abort
+    ${EndIf}
+    DeleteRegKey HKCU "$4"
+    DeleteRegKey /ifempty HKCU "Software\${APP_ORG}-upgrade-backups"
+  ${EndIf}
+  ${If} $6 != 0
+    MessageBox MB_OK|MB_ICONSTOP "The previous uninstaller failed. Your saved settings have been preserved. Please retry setup." /SD IDOK
+    SetErrorLevel 1
+    Abort
```
owenpkent: [P2] Preserve a usable keyboard when legacy cleanup fails

On the same-directory update path, this helper runs before extracting the replacement application. The installer has already closed Alpha-OSK. If the old uninstaller cannot start or returns nonzero, settings restoration completes but this `Abort` prevents the new files from being installed. Previously the caller continued with extraction after `ExecWait`.

This also conflicts with automatic recovery: the detached relauncher waits for a newly installed executable, so an intact old binary does not satisfy its modification-time condition; a partially successful old uninstall can leave no usable binary at all. The user loses access to the keyboard and a retry encounters the same failed cleanup.

Once settings have been successfully restored, continue installing the replacement despite the old uninstaller's failure, or implement an explicit rollback/relaunch path. Keep backup/restore failures fail-closed. Test the generated same-directory install sequence together with the relaunch behavior, since the current helper-only tests intentionally expect the abort and cannot detect this integration failure.

owenpkent (commented):
## Follow-up review of `cd3198a`

The settings-copy helper has useful failure coverage, but its failure contract needs to be checked against the updater. The inline comment describes a reachable same-directory recovery problem.

### Verification

- Ran `tests/test_upgrade_settings.py` and `tests/test_windows_installer.py` against an isolated archive of this head: **70 passed**. The executable registry tests used private UUID-named HKCU keys.
- Checked the worker dependencies: `js-yaml` resolves to 4.3.2 and `sharp` to 0.35.4; TypeScript and a native Sharp image operation passed, Wrangler started, and a full npm audit reported zero vulnerabilities.
- The CI run associated with this head passed. GitHub currently reports a merge conflict, so the resolved commit still needs its checks.

### Corrections to the earlier review

The claimed deletion of the new Add/Remove Programs entry is not established by the actual shipped upgrade path. `build.py` writes the new HKCU `InstallLocation` before `customInstall` reads it, so its different-directory HKCU guard sees `$INSTDIR` and skips cleanup. That ordering is a pre-existing detection problem, and it also prevents the claimed deletion sequence from running. A legacy HKLM branch exists, but the shipped versions inspected use HKCU; its hypothetical behavior should not be presented as a reproduced shipped regression.

Likewise, credential elevation makes both the helper and the old uninstaller use the same administrator HKCU. This does not establish the earlier claim that this change deletes the original standard user's settings. The account-scope question deserves a separate installed-upgrade test.

### Remaining coverage and recovery limits

The executable tests call `RunPreviousUninstaller` directly. Add a harness that exercises the generated installer ordering and the relauncher contract, especially an old uninstaller that fails to start and one that removes files before returning failure.

For restoration failure during a silent update, the recovery registry path is only emitted through `DetailPrint` and a message with `/SD IDOK`. Preserve that path in a discoverable durable diagnostic so the user can actually use the retained backup.

The pytest timeout is a per-test-protocol watchdog. Collection, `pytest_sessionfinish` work such as coverage finalization, and worker shutdown need a process-level timeout if those are the stalls being targeted. The current options are valid, but passing tests do not demonstrate coverage of those phases.

owenpkent:
## Addressed in 6f7061f

Rebased onto current `main`, which is a bigger change than usual: main has since rewritten the previous-version handling (`removePreviousInstallAt`), so the guard now wraps **that macro's** invocation rather than the two copies in `customInstall`. The command it runs is the macro's own validated `$8\uninstall.exe`, never a string read from the registry. The branch's `fix(deps)` commit was dropped, since main carries that fix.

**Two findings are resolved by main rather than by me, and I checked both rather than assuming.** The ARP entry now lives in HKLM and `customInstall` runs *before* the Install section writes it (build.py 996 against 1005), so finding 1 does not arise, matching the follow-up review's correction. Main also added `Delete "$8\uninstall.exe"` and `RMDir "$8"` after the call, which is finding 2.

- **Finding 3 / the inline P2, the abort.** The old uninstaller's exit code no longer stops setup; backup and restore failures still do. That asymmetry is now the thing the code, the docs and a test all say out loud: on the same-directory path the guard runs before extraction, so aborting there left the old install removed, no new one, no shortcuts, and a retry failing at the same point, with the relauncher unable to help because it waits for a newly written executable.
- **Finding 5, the silent path.** A retained recovery copy is recorded in two durable places: a `LastRetainedBackup` value on the fixed `HKCU\Software\alpha-osk-upgrade-backups` key, and `settings-recovery.txt` in the install folder. `/SD IDOK` and `DetailPrint` are both silent on the auto-updater's path.
- **Finding 6, tests that cannot fail.** New pair, `test_the_guard_waits_for_a_real_uninstaller` and `test_without_in_place_execution_the_restore_lands_too_early`: they compile a **genuine NSIS uninstaller** (an installer that runs `WriteUninstaller`, whose uninstall section sleeps then deletes) and drive it with and without `_?=`. I verified the positive half fails when pointed at the detached command, so a dropped `_?=` now fails the suite. The uncovered cell is covered too: `test_a_failing_uninstaller_with_nothing_to_preserve_still_continues`, which is where the old code aborted while claiming settings had been preserved that never existed.
- **Finding 4, credential elevation.** Not silently claimed away. `docs/build/WINDOWS.md`, CLAUDE.md and the changelog now say the elevated installer's `HKEY_CURRENT_USER` is the profile that elevated, so under credential elevation the tree protected is the administrator's. The follow-up is right that it needs an installed-upgrade test; that is not something these helper-level tests can reach.
- **Smaller items.** Include guard added. Both leaking backup-failure paths now clean the parent key. A name already in use is stepped past rather than aborted, since a collision after a temp clean bricked every retry. The `Sleep 1500` is gone (the self-copy window `_?=` removed). `AGENTS.md` is a one-line pointer at `CLAUDE.md` rather than a 2,492-line copy with nothing keeping the two in step. The CI comment now states what the pytest watchdog does **not** cover: it is disarmed during collection, session-scoped teardown, coverage finalisation and xdist shutdown, which is where a "stalls after nearly all tests finish" symptom points, so a green run is not proof.

Full gate green locally (ruff, format, mypy both platforms, pytest), including the Windows NSIS tests with `makensis` present.

PR #108, opened 2026-09-10
fix: improve short-word suggestions and personal learning

owenpkent:
## Summary

- Sparse two-letter prefixes such as `ww` now offer `we` alongside exact completions when the prediction bar has spare slots.

- Shipped training text becomes an in-memory prior instead of being added to personal history at every launch. Real typing gains influence sooner, and confirmed prediction edits immediately teach each word, its context, and its fuzzy entry.

- Pytest and the KSR benchmark clean up generated temporary directories after runs, including failures. Explicit benchmark model directories are preserved. Generated caches, temporary test folders, and installer smoke artifacts are also ignored by Git.

## Related issue

No linked issue. Addresses missing short-word corrections and slow personal learning.

## Type of change

- [x] Bug fix

- [ ] New feature

- [ ] Docs only

- [x] Build / CI / tooling

- [ ] Other (describe)

## Test plan

- [x] `python check.py` passes locally through the pre-push hook: lint, formatting, mypy for Linux and Windows, and the full pytest suite.

- [x] Added or updated tests for short-prefix rescue, explicit edit learning, privacy/freeze guards, corpus persistence/reload/clear behavior, and temporary-directory cleanup.

- [x] Verified on Windows through automated tests and benchmark runs.

- [ ] Verified on Linux

- [ ] Verified on macOS

Fresh-model learning checks with five pills and a one-letter prefix put "bamboo" first after 3 full-word learns instead of 22, or 1 selection instead of 5. These are engine-level examples, not guarantees for mature personal history.

Five-pill KSR after the changes: `aac-dev` 49.1% clean / 45.5% with a mis-click per word; `aac-test` 50.4% / 46.9%. Clean scores were unchanged; the small slip gains are below the documented noise threshold. A real KSR run also confirmed automatic removal of its temporary model folder.

The later ignore-only commit was verified with `git check-ignore`, including inverse checks that source files, tests, build scripts, documentation, and training data remain eligible for tracking.

## Accessibility check

- [ ] Works with default repeat-delay / repeat-interval settings

- [ ] Visual feedback still works for low-vision users

Timing and visual feedback are outside this change. No desktop input was synthesized during verification.

## Notes for reviewers

This changes the prediction engine. The corpus prior uses weight 0.1, shared by n-gram and fuzzy scoring. Existing saved history is preserved because historical corpus counts cannot safely be separated from real typing; this makes adoption gradual for existing users.

Backup reload rebuilds the prior against the replacement history, so words admitted only by the previous history cannot linger in suggestions. Confirmed edits bypass the unknown-word gate; ordinary prediction selections retain it, and privacy and learning-freeze guards remain enforced.

The base dictionary data, default rank merge, and export schema are unchanged. Prediction documentation records the measurements and limits.

The worker development dependency overrides now require `js-yaml` 4.3.2 and `sharp` 0.35.4, with the corresponding lockfile updates. OSV reports no issues in the npm lockfile; a clean `npm ci`, `npm audit`, TypeScript check, SBOM generation, and Wrangler build dry run passed. Telemetry code, payloads, and consent behavior are unchanged.


owenpkent:
## Review

The short-prefix rescue and the corpus-prior split are both good changes and the test coverage on the rescue is solid. One functional bug and a few consistency issues.

### 1. An unmodified prediction edit permanently injects a never-typed word

`editPrediction` (src/keyboard_bridge.py:5554) never compares `edited` to `original`:

```python
edited = self._sanitize_edit(edited)
if not edited:
    return
if not self._privacy_mode:
    self._predictor.learn_from_selection(self._context_buffer, edited, explicit=True)
```

So right-click a pill, open Edit, tap Save without changing anything, and the new `explicit=True` path runs. `learn_word` writes +5 into `user_vocab`, `unigrams` and `_user_total`, and `reinforce_context` writes a persisted user bigram edge.

The failure case is a pill the fuzzy source or prefix beam generated for a word the user has never typed. One tap and it is permanently in the model, which is precisely what `learn_from_pill_click`'s three-sighting gate exists to prevent, and which this PR deliberately preserves for ordinary pill clicks. Opening the editor to look at a word and closing it via Save is an ordinary thing to do.

Guarding the explicit branch on `edited != original` closes it, and keeps the intended meaning: explicit means the user corrected something.

### 2. The explicit path bypasses the shape filter and the blacklist

`learn_from_selection(..., explicit=True)` calls `NgramPredictor.learn_word` directly (src/prediction/hybrid_predictor.py:993), which applies neither `_is_plausible_word` nor `is_suppressed`. Every sibling entry point (`learn`, `learn_from_pill_click`) applies at least the shape filter as defence in depth.

Two consequences:

- Edit a pill to `zzz` (no vowel) or `pr` (two letters, not on the profile's `short_words`) and it lands in `user_vocab` at 5. `load()`'s fragment strip removes it from `unigrams`/`user_vocab` on the next launch, but the `_user_bigrams` edge `reinforce_context` wrote is persisted and is never stripped, so a corrupt edge survives restarts.
- The same path re-adds a word the user explicitly blacklisted, since `learn_word` has no suppression check.

Routing the explicit path through the same filters, or at minimum through `_is_plausible_word` and `is_suppressed`, would match the rest of the file. Note the deliberate exception for taught acronyms if you do (`is_taught_acronym`).

### 3. `_install_corpus_unigrams` persists what its own comment says it does not

The comment reads "It stays in memory only, so it cannot inflate the persisted user history" and the doc table says "Rebuilt in memory, never saved as personal history". But `_install_corpus_unigrams` (src/prediction/ngram_predictor.py:1049) writes into `self.unigrams` and `self.total_words`, and `save()` persists both.

Two effects:

- A corpus-only word gets written to `ngram_model.json` on the first save. The function only ever raises (`if current < desired`) and nothing subtracts, so if a later release drops that word from `data/training_corpus.txt` the stale entry stays forever, keeps passing `_is_valid_word`, and keeps being fed to the prefix index by `_fuzzy_frequencies`.
- The count lands at full weight rather than `_CORPUS_PRIOR_WEIGHT` (0.1x) in the two places that read `unigrams` directly: `_top_unigrams_with_scores` (the no-context ranking path) and `_fuzzy_frequency`'s `base = _base_unigrams.get(word, 0) or unigrams.get(word, 0)` fallback, which for a corpus-only word returns the full corpus count as its "base".

If the intent is genuinely memory-only, `_base_unigrams` looks like the right home. If persistence is intended, the comment and the doc table need to say so.

### 4. Raising the max-suggestions setting can reorder the leading pills

Whether the short-prefix rescue fires is a step function of `n`. This PR's own tests show both sides: `predict("ww", n=2) == ["wwe", "wwii"]` (no rescue, 2 passing >= 2) but `predict("of ww", n=5)` puts `we` inside the top 3.

In `_score_rank` an n-gram rank-1 pill scores 1.0 while a fuzzy rank-1 pill scores `0.6 * (1 + log1p(bigram_count)/2)`, which exceeds 1.0 for any bigram count >= 3, and curated seeds are +50. So a one-substitution neighbour with context support takes pill slot 1 from the exact completion.

A user changing Settings, Smart Typing, Suggestions, max count from 3 to 5 therefore finds the word in slot 1 has changed for short prefixes, on a keyboard where pill position is muscle memory. The docs claim only that "prefixes with enough exact suggestions keep their existing ranking".

`test_sparse_real_prefix_keeps_its_exact_prediction` only covers the no-context case, where the bonus is 1.0, so it cannot see this. Worth either deciding the rescue on a fixed threshold rather than on `n`, or documenting the interaction and adding the with-context case to that test.

### 5. Smaller items

- **`include_user_candidates` is now dead.** Both call sites (`_reseed_context`, `_load_training_corpus`) pass `False`, so the `True` branch reading `_candidate_counts` is untested dead code. CLAUDE.md's context-table section still says the reseed "applies the same known-or-third-sighting gate as `learn(corpus=True)` (reading the candidate pool, never writing it)", which is now false at both call sites. Either drop the parameter or update that sentence.
- **`ensure_corpus_prior` is a pure alias for `load_corpus_prior`.** Two public names for one operation, and the names imply a distinction ("ensure" as idempotent restore, "load" as rebuild) that does not exist, so a future change to one path silently applies to the other.
- **The explicit path leaves a stale candidate entry.** `learn_word` promotes into `user_vocab` without popping `_candidate_counts` / `_candidate_last_seen`, unlike `learn_from_pill_click`'s promotion which pops both. Every later `learn()` then takes the known-word branch and never reaches the pop, so the entry is persisted under `candidate_counts` until decay or the age sweep retires it.
- **`docs/architecture/HYBRID_MERGING.md:312` points at `AGENTS.md`**, which does not exist on `main`. Three other lines in the same file (19, 710, 771) reference `CLAUDE.md`. Note #107 adds an `AGENTS.md` as a byte-identical copy of `CLAUDE.md`, so merge order decides whether this resolves, and either way two authoritative maps is the thing CLAUDE.md warns about.
- **No CHANGELOG entry for the prediction changes.** The only `[Unreleased]` addition covers the worker's npm bump. The corpus prior, the short-prefix rescue and the explicit edit learning all change what the user sees, so the release notes would carry no record that the engine moved.
- **`tmp_path_retention_policy = "none"`** deletes failed tests' temp directories too, which is the evidence a failure needs. pytest's `"failed"` achieves the stated cleanup goal and keeps post-mortem data for CI artifact upload.
- **`_candidate_passes` runs twice per n-gram candidate per keystroke**, once in `predict()` to compute `allow_short_prefix` and again in `_merge_predictions` via `_score_rank`/`_normalise_source`. On the next-word path the flag is computed and passed down but `get_fuzzy_predictions` returns `[]` immediately because `current_word` is empty, so the whole count is discarded. Computing it lazily only when the current word is 1 to 2 characters would skip both.

src/prediction/ngram_predictor.py
```diff
@@ -1023,6 +1046,50 @@ def learn_corpus_context(self, text: str) -> None:
                 learned.append(word if sightings[word] >= self._candidate_threshold else None)
         self._link_context(learned, base=True)
 
+    def _install_corpus_unigrams(self) -> None:
+        """Make corpus words visible in the merged vocabulary without user counts."""
+        for word, count in self._corpus_unigrams.items():
+            desired = self._base_unigrams.get(word, 0) + self.user_vocab.get(word, 0) + count
+            current = self.unigrams.get(word, 0)
+            if current < desired:
+                self.unigrams[word] = desired
```
owenpkent: [P2] Keep corpus-only vocabulary from surviving a removed corpus

This assignment feeds a table that `save()` persists. Reproduction on this head: load `zzqcorpus` three times only through `load_corpus_prior`, verify its personal count is zero, then save. The JSON contains `unigrams.zzqcorpus == 3`. Restart with the training corpus empty: `_corpus_unigrams` is empty, but the merged count is still 3, `_is_valid_word("zzqcorpus")` is true, and the real hybrid rebuilds the word into its fuzzy dictionary.

Thus a future corpus correction/removal cannot retire its old vocabulary after the first save, despite the new prior's intended rebuild-from-shipped-data lifecycle. This confirms the earlier review's persistence concern independently of user-count inflation.

Track corpus provenance outside the durable merged vocabulary, or account for that contribution during serialization/rebuild without subtracting actual personal learning. Add a save/restart regression where a corpus-only token is removed from the next corpus and a genuinely learned token is preserved.

owenpkent (commented):
## Follow-up review of `87f519a`

Revalidated the corpus-persistence concern from the earlier review with a save/restart probe against this exact commit. Corpus-only vocabulary remains in the saved merged table and returns to the fuzzy dictionary even when the replacement corpus no longer contains it. The inline comment records the result and a regression test to add.

### What passed

**56 focused tests passed**, covering `test_corpus_prior.py`, hybrid learning, sparse two-letter rescue, fuzzy completion, and prediction-edit privacy. Tests ran in an isolated source archive with temporary model directories. The CI run associated with this head also passed.

Source inspection confirmed that `_user_total` continues to track `sum(user_vocab.values())`, and that the learning-freeze and privacy checks remain in place around explicit edit learning. The short-prefix rescue continues to defer to a sufficiently populated valid exact-candidate set.

### Clarify the explicit-edit contract

An unchanged value saved through the editor does take the new +5 path. Whether that is wrong depends on whether Save confirms the displayed word or only a textual change counts as correction; the new API documentation supports the former interpretation. I would resolve that behavior explicitly in a test and the documentation rather than assume the earlier review's proposed inequality guard is required.

Separately, explicit edits can admit a token that normal load-time plausibility filtering later rejects. Add a save/reload test for such an edit and for an explicitly suppressed word, then keep the learning and reload rules consistent. For a user who deliberately teaches an acronym or personal token, silently forgetting it on restart is a different outcome from rejecting it at edit time.

### Merge readiness

GitHub currently reports a merge conflict. After resolution, rerun the focused lifecycle tests and the full gate. Keep the corpus-weight measurements tied to the named AAC split and retain the distinction between fresh-model gains and behavior with existing saved history. No live model, desktop insertion, or benchmark run was used in this follow-up.

owenpkent:
## Addressed in 1982333

Rebased onto current `main`. The branch's `fix(deps)` commit was dropped: main already carries the `js-yaml` 4.3.2 / `sharp` 0.35.4 fix with a fuller changelog entry, and the package files were identical.

- **Corpus prior persisted (finding 3, inline P2):** `_install_corpus_unigrams` is gone. The prior lives in `_corpus_unigrams` only and never touches `unigrams`, so `save()` never sees it. Readers go through two new methods, `NgramPredictor.in_vocabulary` (membership, used by `_is_valid_word`) and `vocabulary()` (enumeration, used by the fuzzy dictionary rebuild); `_top_unigrams_with_scores` adds the prior at `_CORPUS_PRIOR_WEIGHT` rather than at full count, and `_fuzzy_frequency`'s fallback no longer sees a corpus-only word as base. Regression: `test_a_word_dropped_from_the_corpus_leaves_with_it` (corpus-only token removed from the next corpus is gone from candidates and the fuzzy dictionary; a `learn_word`-taught token stays), plus `test_a_corpus_word_is_never_written_to_the_model_file`, which reads the saved JSON. `in_vocabulary` also consults `_base_unigrams`, which closed the one test that broke: `load()` replaces the merged table with the file's, so the poisoned-model hardening test's `hello` was only in the base table; that test now checks the prior itself.
- **Unchanged Save (finding 1, follow-up):** resolved explicitly rather than assumed. `editPrediction` passes `explicit=edited.lower() != original.lower()`: a changed spelling is a correction and learns at once; an unchanged or casing-only Save is one ordinary pill tap (gated), with casing still recorded. Tests: `test_an_unchanged_save_is_one_ordinary_tap`, `test_a_casing_only_save_is_an_ordinary_tap_that_records_the_casing`, `test_a_changed_spelling_is_explicit`. Docstrings and CLAUDE.md say why.
- **Explicit path bypassed the filters (finding 2, follow-up):** the explicit branch applies `_is_plausible_word` (taught acronyms exempt as everywhere) and `is_suppressed`, and a refused token breaks the context chain so no `_user_bigrams` edge spans it. Tests: `test_explicit_selection_refuses_what_load_would_strip` (`zzz`, `pr`: nothing in `user_vocab`, no persisted edge, not in the fuzzy dictionary), `test_explicit_selection_refuses_a_removed_word`, `test_a_refused_token_breaks_the_context_chain`, and `test_an_explicit_edit_survives_a_save_and_reload` for the accepted side, so the two rules agree.
- **Rescue depended on `n` (finding 4):** decided against `HybridPredictor.SHORT_PREFIX_RESCUE_FLOOR` (5, the benchmark's pill count, so the measured figures describe this rule) instead of the requested count, and only computed for a two-character current word. When `n` is small the n-gram is asked for enough to decide, so a plentiful prefix cannot look sparse. Tests: `test_the_rescue_does_not_depend_on_how_many_pills_are_shown` (every bar is a prefix of the widest one, for `ww` / `of ww` / `he` / `the he`) and `test_a_small_pill_count_cannot_make_a_plentiful_prefix_look_sparse`.
- **Smaller items:** `include_user_candidates` removed (local sightings only at both call sites; CLAUDE.md's context-table sentence updated), `ensure_corpus_prior` removed, `learn_word` pops the candidate entry (`TestLearnWordRetiresTheCandidateEntry`), `AGENTS.md` -> `CLAUDE.md`, three CHANGELOG entries under Unreleased, `tmp_path_retention_policy = "failed"`, and `_candidate_passes` runs once, only where the rescue can apply.

Full gate green locally (ruff, format, mypy both platforms, pytest).

PR #109, opened 2026-09-11
docs: record architecture review and recommendations (superseded by #129)

owenpkent:
> **Superseded by #129, which merged as `6a1a6de`. Nothing here was lost.**
>
> This PR was never closed by a decision. It was stacked on #108, and merging that with `--delete-branch` deleted the base ref, so GitHub retargeted this to `main` and closed it in the same second (`automatic_base_change_succeeded` / `base_ref_deleted` / `closed`, all at 2026-09-16T03:28:46Z). It cannot be reopened or merged: the branch was rebased after the close, so the recorded head `b198933` is no longer in its history and GitHub answers `Pull Request is not mergeable`.
>
> The same branch and the same content, rebased onto `main`, went in as #129.

---

## Summary

- Record the architecture at commit 87f519a, with a system diagram, current measurements, strengths, and prioritized recommendations.
- Cover operation failure contracts, transactional restore, interaction and settings ownership, responsiveness, and experimental prediction lifecycle.
- Link the dated review from the documentation index. Implementation remains follow-up work.

## Type of change

- [x] Docs only

## Accessibility check

N/a: this PR changes documentation only.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,6 +20,7 @@ Active design docs for the engine, the platform abstraction, and the user-facing
 - [`architecture/FUZZY_RECOGNITION.md`](architecture/FUZZY_RECOGNITION.md): spatial error correction, tunable constants, the relationship to SymSpell.
 - [`architecture/PREDICTION_NOTES.md`](architecture/PREDICTION_NOTES.md): the "unified system" framing across the predictors, fragment filter and repetition gate, autocorrect thresholds, reinforcement and backspace-as-negative-signal, the structured-token store for numbers and email addresses, prioritised future work.
 - [`architecture/PLATFORM_ARCHITECTURE.md`](architecture/PLATFORM_ARCHITECTURE.md): cross-platform abstraction details (key synthesis, password-field detection, config paths).
+- [`architecture/ARCHITECTURE_REVIEW_2026-09-10.md`](architecture/ARCHITECTURE_REVIEW_2026-09-10.md): dated architecture assessment and prioritized recommendations for failure handling, data restore, interaction ownership, settings and responsiveness.
 - [`architecture/COMPACT_VIEW.md`](architecture/COMPACT_VIEW.md): the denser 13x4 keyboard for small screens, its measurements, and the layer model behind `?123`.
 - [`architecture/MODULAR_LAYOUTS.md`](architecture/MODULAR_LAYOUTS.md): custom keyboard layouts inspired by Octavium / Nimbus.
 - [`architecture/DICTATION.md`](architecture/DICTATION.md): voice input. The Qt-only audio and websocket stack, the four-state run model, where the Deepgram API key lives, and why privacy mode cancels a run outright.
--- a/docs/architecture/ARCHITECTURE_REVIEW_2026-09-10.md
+++ b/docs/architecture/ARCHITECTURE_REVIEW_2026-09-10.md
@@ -1,0 +1,216 @@
+# Architecture review, 10 September 2026
+
+Reviewed the clean working tree at `87f519a`. This is a dated assessment,
+not a description of changes already implemented. Recommendations below
+are proposals; the review does not change runtime behavior.
+
+**Keep the Python/PySide6 architecture and improve its boundaries in small
+steps.** The platform adapters, local prediction engines and QML components
+are useful separations. The main weaknesses are concentrated in operation
+failure handling and the ownership of interaction state. Those matter
+especially here: an incorrect success assumption can affect the next text
+replacement, and an unresponsive keyboard removes the user's input device.
+
+```mermaid
+flowchart TD
+    UI[QML keyboard, editors and settings] --> B[KeyboardBridge]
+    B --> S[Platform input synthesis]
+    S --> T[Focused application]
+    T -. focus, caret and password observation .-> B
+    B <--> P[Local n-gram, fuzzy and token prediction]
+    B --> D[Models, snippets and analytics stores]
+    B <--> V[Optional dictation controller]
+    UI --> F[Separate telemetry and study bridges]
+```
+
+The feedback edge is fundamental. The keyboard reconstructs context in
+another application without owning its text buffer. It therefore needs
+explicit rules for when that reconstructed state is trustworthy.
+
+The current bridge is 5,835 lines; `Main.qml` is 5,192 and
+`UnifiedSettingsPanel.qml` is 3,092. These are physical line counts including
+comments and blank lines, not complexity scores. They locate the
+concentration of responsibilities rather than establish a reason to rewrite.
+
+There are strong foundations to preserve: centralized modifier and verbatim
+insertion helpers, explicit window nonactivation, privacy-triggered context
+scrubbing and dictation cancellation, separate base and personal prediction
+data, hardened archive allow-lists, shared atomic file replacement, and real
+QML tests alongside property tests. CI runs tests on Windows and Linux and
+checks Python types for both platforms. The earlier structural review's
+floating-window extraction, atomic writes, shared pack-ID rules, telemetry
+extraction and platform-window extraction have already landed. The C++ branch
+is explicitly parked.
+
+1. **Immediate: extend the privacy boundary through platform diagnostics.**
+
+   In [`linux.py`](../../src/platform/linux.py), `_run` logs the complete
+   subprocess command on timeout or exception. `send_text` puts input text
+   into that command. Privacy mode intentionally permits typing, so it
+   cannot prevent these error logs from containing password characters or
+   inserted phrases.
+
+   An isolated probe replaced `subprocess.run` with a timeout and captured
+   the logger call. The rendered error contained the synthetic input text.
+   No real input was sent.
+
+   Define diagnostics at the adapter boundary as backend name, operation
+   type, duration and safe error code. Exclude payloads and exception strings
+   that can contain command arguments. Add timeout and exception regressions
+   that assert a synthetic secret never reaches logging. This is a confirmed
+   defect with a small, independently shippable fix.
+
+2. **Immediate, then incremental: give operations honest success and failure contracts.**
+
+   Three boundaries currently lose information their callers need:
+
+   - Input: [`KeySynthesizerBase`](../../src/platform/base.py) returns `None`.
+     Linux ignores nonzero subprocess exit status and swallows timeouts;
+     Windows `_inject` logs incomplete `SendInput` submission without
+     reporting it upstream. `_press_char` then advances the bridge context.
+     A later completion can therefore operate on text that was never sent.
+   - Persistence: [`SnippetStore.save`](../../src/snippets.py) catches
+     `OSError`; mutators update memory and return `True` anyway. A mocked
+     write failure confirmed that `set` reports success and changes the live
+     value, even though persistence failed.
+   - Release: [`build/windows/build.py`](../../build/windows/build.py)
+     `main` ignores the Boolean results of `sign_installer` and
+     `verify_build`. A fully mocked build with both returning `False` still
+     returned exit code 0 and printed that all outputs were signed. The
+     updater independently verifies downloads, so this finding establishes
+     misleading build success, not an updater signature bypass.
+
+   First make failed installer signing or verification fail the release
+   command, and make snippet persistence failure reach the editor. Keep
+   intentional unsigned development builds explicit. Then introduce a small
+   input result distinguishing submitted, rejected and partial/uncertain.
+   On known failure, stop dependent edits, invalidate uncertain context and
+   suppress associated learning and success metrics. Never automatically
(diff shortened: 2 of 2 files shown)


owenpkent:
## Review

Useful assessment, and the diagnostics item is a real find. The issues are about durability of the record and about two places where it contradicts the existing architecture docs without saying so.

### 1. The anchor commit will not exist after #108 merges

Line 3 anchors the review to commit `87f519a`. That commit is the head of open PR #108:

- `git merge-base --is-ancestor 87f519a origin/main` returns false
- `git branch -r --contains 87f519a` lists only `origin/fix/prediction-learning-and-temp-cleanup` and this branch

This repo squash-merges, so once #108 lands that SHA is unreachable and `git show 87f519a` on a fresh clone gives `unknown revision`.

Two further consequences worth stating in the document either way: the tree described is `main` plus roughly 985 lines of unmerged prediction-engine work (`hybrid_predictor.py`, `ngram_predictor.py`, `prefix_beam.py`, `fuzzy_recognizer.py`), which the document does not mention, so the quoted line counts (5,835 / 5,192 / 3,092) describe a state that never shipped.

`STRUCTURAL_REVIEW.md` anchors to a tag instead ("taken at **v1.2.2** against `main`"), which is the durable form.

### 2. The confirmed log leak ships with no fix and no purge step

Item 1 is correct and it is the most valuable thing in the document. `src/platform/linux.py:130` and `:132` interpolate the whole `cmd` at ERROR, and `send_text` builds `["xdotool", "type", "--clearmodifiers", "--", text]`, so on any Linux host where xdotool stalls, `alpha-osk.log` records the characters typed, including into a password field. CLAUDE.md makes this a hard invariant and it is the file bug reporters attach.

Two gaps in the remediation as written:

- It names only `_run`. The same file has `_logger.warning("No synth tool - cannot send key: %s", key_name)` at :223, and `key_name` is a typed character on the chord path (keyboard_bridge.py:1624) and the game-hold path (:1637). On a host with neither xdotool nor ydotool installed that fires on every `send_key`. A fix built from this text redacts one site and leaves the other.
- It omits the project's established second half. `keyboard_app.py::_purge_pre_fix_logs` and the `.log-privacy-purge` sentinel exist because "fixing the logging sites only stops new leakage; an upgrading user still had up to four rotated files holding a transcript of what they typed". Implementing only lines 57 to 61 leaves every already-leaked transcript on disk.

Given it is a confirmed, live privacy defect, it may be worth pulling item 1 out of this document into its own PR rather than landing it as a recommendation.

### 3. Two recommendations contradict `STRUCTURAL_REVIEW.md` without acknowledging it

- Recommendation 2 (lines 85 to 89) asks to "introduce a small input result distinguishing submitted, rejected and partial/uncertain. On known failure, stop dependent edits, invalidate uncertain context and suppress associated learning and success metrics." That is a change to `_press_char` / `_send_key` / `_send_text` and the context buffers. `STRUCTURAL_REVIEW.md:356` states the opposite as a governing constraint: "Nothing here requires touching the keystroke state machine, which is the part that is hard, correct, and best left alone."
- Recommendation 4 (line 126) "Begin with constructor injection" runs against STRUCTURAL_REVIEW section 3.1, "The bridge cannot be refactored at today's price, and the tests are why", and its step 6 conclusion "This is a measurement, not a commitment."

Both may well be the right call now, but two documents in `docs/architecture/` giving opposite guidance on the most invariant-dense code in the project, with nothing saying which supersedes which, is worse than either position alone. A sentence naming the earlier constraint and saying why it is being revisited would settle it.

### 4. Filing location contradicts the docs README's own rule

`docs/README.md:83-86` defines the split: "**`architecture/`**: a live system that's currently shipping. The code in `src/` should match what the doc says" and "**`research/`**: written to inform a decision (or capture an audit) but not the source of truth for current behaviour."

This document's own first paragraph disqualifies it from the first bucket ("This is a dated assessment, not a description of changes already implemented. Recommendations below are proposals"), and its content is a list of things `src/` does not do. The section header the new index entry sits under adds "Edit when the code in that area changes", which is the opposite of what a dated snapshot wants. `research/SECURITY_AUDIT.md` is the existing precedent for this exact shape.

### 5. Smaller items

- **No headings.** 216 lines with exactly one heading, so no recommendation has a markdown anchor. The sibling docs have 18 (`STRUCTURAL_REVIEW.md`) and 25 (`research/SECURITY_AUDIT.md`), and the repo cites them by section ("sequence item 7 of...", "per ... section 3.1"). Nothing can deep-link to "the diagnostics redaction item".
- **`STRUCTURAL_REVIEW.md` is referenced in bare prose at line 40 with no link**, while all ten source files are hyperlinked, and it is absent from `docs/README.md` entirely. So the index this PR edits lists the new review and not the one it continues. A reader arriving via the index cannot find that the earlier doc carries a live status table and the measured `Main.qml` figure (6,004 down to 4,338 after #55) that makes 5,192 a regrowth rather than a bare number.
- **The `verify_build` claim is unqualified but only half true.** build/windows/build.py:1217-1224 is the `--verify-only` branch and already does `ok = verify_build(signtool); return 0 if ok else 1`. Only the full-build path (1271, 1275) discards both results. Someone auditing the claim against `--verify-only` finds it false and may discount the whole item, which is a real defect on the path that matters.
- **Recommendation 8 reverses a documented decision.** "Require a successful rescue before destructive replacement" contradicts `src/data_export.py:437-439`, which documents the current behaviour as a choice: "Rescue export failures are logged but do not abort the import." Making it a precondition means a user whose disk cannot hold a second full copy can no longer restore their backup at all, which is the worse direction for the one artifact CLAUDE.md calls irreplaceable. Combined with line 106 ("Stage and validate all payloads before changing live files") the import would need roughly three times the dataset in free space. Neither cost is mentioned.
- **The diagnostics recommendation is stated as a blanket rule** ("exclude payloads at the adapter boundary") but the platform layer deliberately logs payloads at DEBUG and `keyboard_app.py:684-691` records that as policy: "DEBUG is the sanctioned place for typed content precisely because it is off in normal operation." Distinguishing the ERROR site (the actual defect) from the DEBUG sites (already compliant) would stop a reader stripping a deliberate affordance.
- **The window-focus inventory misses `helpWindow`.** `qml/Main.qml:4881` (settings), `:5137` (help) and `:5169` (viz) all declare the same flag set with no `WindowDoesNotAcceptFocus`, and `keyboard_app.py:290-294` wires only snippets, symbols, study and viz. Help is a window the user opens while typing into another app, so it is the one most worth a recorded decision.
- **The system diagram draws `UI --> B` one-way**, omitting the Python to QML signal edge that is half the documented bridge pattern and the exact path recommendation 4 depends on (edit-mode signal routing, lines 133 to 136).
- **Recommendation 7 asks for a decision CLAUDE.md already records.** The PPM section states the switch (`_ppm_in_merge`, default False), why training is retained (a later fusion inside the prefix beam, the VelociTap shape) and the measured merge cost (21ms to 3ms per keystroke). The genuinely unmeasured part is the *training* cost, so asking only for that would be the actionable form.
- **Index entry wording.** It lists recommendations 2 through 6 but omits recommendation 1, the confirmed privacy defect and the only independently shippable item, so a reader scanning the index for logging work never opens the file. Also `docs/README.md:22` reads "prioritised" and the new line 23 reads "prioritized", one line apart.

owenpkent (commented):
## Follow-up review of `2a5e017`

Reviewed this as the two-file documentation change against its actual base, `87f519a` on #108. No additional correctness finding beyond the existing discussion.

### Evidence checked

- Recounted all three quoted files at the named commit: `keyboard_bridge.py` 5,835 lines, `Main.qml` 5,192, and `UnifiedSettingsPanel.qml` 3,092. The measurements match.
- All 11 local links in the assessment resolve in that commit's tree.
- Traced the Linux timeout/exception logging, snippet mutation/save behavior, and full-build signing/verification calls. The described failure paths are present. The build finding correctly concerns misleading release-command success; it does not establish a bypass of the updater's signature verification.
- The document explicitly separates reproduced failures from the unproven overlapping-editor risk and from proposed architectural changes. That distinction should survive editing.
- The CI run associated with this head passed. This follow-up used source inspection and link/count checks; it did not rerun the full suite or the original mocked probes.

### Suggested resolution of the existing discussion

Prioritize the documentation's status and implementation boundaries: identify the reviewed tree as #108's unmerged state, retain a discoverable PR reference when the branch is squash-merged, and make the dated assessment distinct from the live architecture contract. A source SHA remains a useful exact identifier, but an ordinary fresh clone is not guaranteed to contain a discarded branch's objects.

For the failure-contract proposal, explicitly retain the staged approach: fix demonstrable false-success reports, then introduce input failure handling with sequence tests. Likewise, explain the storage-space/recovery tradeoff of requiring a rescue export before restore. These are substantive follow-up design choices and should not be implied to be implemented by this documentation PR.

Keep this PR stacked on #108 until that change lands, then retarget and verify that the final diff still contains only the assessment and its index entry.

owenpkent:
## Addressed in 72d3dcd

Verified every item in both reviews against the tree before editing. All of them held, with two refinements noted at the end.

### Done

1. **Anchor.** The document now identifies the reviewed tree as v1.4.1 (`438cfd3`) plus #108's three unmerged commits, keeps `87f519a` as the exact identifier, and says the SHA will not survive the squash-merge. It also records that the three quoted files differ from v1.4.1 by one line between them, so the counts describe a shipped state to within a line.
2. **Log leak.** Item 1 names both sites (`_run` at ERROR, `send_key`'s missing-tool warning), separates them from the sanctioned DEBUG sites in the prediction path, and adds the sentinel-guarded purge as the second half of the fix. It says the fix should ship as its own PR.
3. **STRUCTURAL_REVIEW.** Items 2 and 4 each name the constraint they revisit (section 6's "leave the state machine alone"; section 3.1 and step 6's "a measurement, not a commitment") and give the reason. The intro states that the earlier review remains the record of its own sequence and status.
4. **Filing.** Moved to `docs/research/`, per the index's own rule and the `SECURITY_AUDIT.md` precedent. The index entry moved with it, now leads with the confirmed defect, and matches its neighbours' spelling.
5. **Smaller items.** A heading per recommendation so each has an anchor. `STRUCTURAL_REVIEW.md` linked in prose and added to the docs index, which had omitted it. `verify_build` qualified to the full-build path (the `--verify-only` branch is correct). The rescue-export tradeoff stated, with the storage cost named and the choice left to the user. `helpWindow` added to the focus inventory. The Python-to-QML signal edge drawn in the diagram. Item 7 reframed around the unmeasured training cost, since the merge decision is already recorded.

### Two refinements to the review comments

- The game-hold path (`_press_char`, the `_in_game_mode` branch) cannot reach the Linux warning: `_window_is_game` takes an hwnd, so `_game_auto_active` never flips off Windows. The document cites the chord path only.
- "Recommendation 8" in the first review is item 3; the document has seven.

### Not done here

The redaction and the purge themselves. That is a code change with tests and belongs in its own PR, as the document now says.

Still stacked on #108. Retarget to `main` after it lands and confirm the final diff is still the assessment and its two index entries only.

owenpkent:
The item 1 fix (redaction of both Linux log sites, the macOS sibling, and the second purge generation for Linux and macOS) is #120.

owenpkent:
## Addressed in b198933

The earlier round (72d3dcd) answered every item in the first review; this pass handles what the follow-up asked for and what has changed since.

- **Still stacked on #108**, now rebased onto its revised head, so this PR's own diff is the assessment and its two index entries and nothing else (`git diff 1982333 HEAD --stat`: `docs/README.md` +2, the document +325).
- **Re-anchored.** The reviewed SHA is perishable for a second reason the document did not know about: #108's branch was revised in review and force-pushed, so `87f519a` is already off that branch and the PR's current head describes a different tree, before any squash-merge. The document now says so and leans the anchor on the **v1.4.1** tag, with the quoted counts stated against it (`keyboard_bridge.py` 5,835 against v1.4.1's 5,834; `Main.qml` and `UnifiedSettingsPanel.qml` identical), which I recounted.
- **Item 1 shipped.** The redaction of both Linux sites, the macOS sibling and the second purge generation landed as #120, and the document records it. A dated assessment that never says which of its findings were acted on is worth less at every later reading; everything else in it stays a proposal.
- **Index entry** picked up main's newer `SECURITY_AUDIT.md` description in the rebase, and its own line now names #120 rather than promising the purge as future work.

The follow-up's other two asks were already in 72d3dcd: the staged approach for the failure contract, and the storage and recovery tradeoff of requiring a rescue export before a restore, both stated as design choices this documentation PR does not implement.

owenpkent:
Closed automatically, not deliberately: this PR was stacked on #108, and merging that with `--delete-branch` deleted the base ref, which retargeted this to `main` and closed it in the same second. GitHub refuses to reopen it because the branch was rebased afterwards and the recorded head `b198933` is no longer in its history. Continued as #129 with the same branch and the same content, rebased onto current `main`.

Issue #106, opened 2026-09-09
[Feature] Announce key and prediction bounds over IPC for external switch scanning


enaboapps:
## Problem

I'm developing Switchify PC into a full switch-scanning system. It needs a keyboard that exposes its interactive targets so Switchify can draw scanning highlights over them and activate the user's selection.

Alpha-OSK would provide the keyboard and predictions. Switchify would manage scanning order, timing, grouping, switch input, and the overlay. Implementing the scanning system itself is outside this issue.

## Proposed feature

Add a documented local IPC interface, initially on Windows, through which an external accessibility application can:

- Discover and connect to a running Alpha-OSK instance.
- Obtain a complete snapshot of visible keyboard keys and prediction pills.
- Subscribe to changes in target geometry, labels, availability, and state.
- Activate a target by its identifier using Alpha-OSK's existing input handlers.

Each target should expose an identifier, displayed label, target type, enabled state, screen-space bounds, and section/row/order information. Modifier targets should also expose their active and locked state.

Bounds must reflect the rendered controls, including compact layouts and optional panels. The coordinate convention must be documented and support mixed-DPI monitors and negative desktop coordinates.

Updates should cover window movement and resizing, monitor/DPI changes, layout and panel changes, modifier changes, prediction replacement, and keyboard visibility. Connecting after launch must provide the current snapshot without requiring a keyboard restart.

Activation should behave like a normal primary click, including sticky modifiers, programmed keys, prediction insertion, and controls that change keyboard pages. It must preserve focus in the application receiving the text.

Targets need identities or revisions that prevent an outdated selection from activating a different prediction or a replaced control.

## Boundaries and alternatives

The first version covers visible keyboard keys and prediction pills. Settings dialogs, snippet and symbol pickers, secondary-click actions, and held-key commands are outside the initial scope.

Keep the interface local and restrict access to authorized clients in the current user session. Publish only the target information needed for the integration, respecting existing privacy behavior. Do not expose typing history or arbitrary code execution.

Screen recognition and hard-coded coordinates would be fragile as layouts, scaling, and predictions change. Windows accessibility APIs are another option to assess, but this proposal requests an explicit, documented integration contract.

Leave transport selection and exact message schemas for implementation discussion. Keep the contract reusable for future platforms.

## Acceptance criteria

- [ ] Switchify or a small test client can connect before or after keyboard startup and enumerate current targets.
[...]

PR #110, opened 2026-09-11
docs: research secure external switch scanning

owenpkent:
## Summary

- explain the normal switch-scanning loop, including automatic, step, group, row-column, point, and radar scanning
- compare built-in platform scanners, AAC products, Windows access tools, and Switchify
- document typical switches, 3.5 mm connections, USB and Bluetooth interfaces, adaptive controllers, and tecla-e
- define a scanner-neutral target contract for Alpha-OSK across UIA, macOS Accessibility, and AT-SPI
- recommend Windows UIA with cached polling before adding any local notification channel

## Security

- distinguish operating-system accessibility authorization from partner-exclusive authentication
- preserve Alpha-OSK's no-focus, privacy, and one-shot activation invariants
- document stale-target handling and the risks of a same-user activation pipe
- bound the controls required if authenticated activation IPC ever becomes necessary

## Validation

- pre-push Ruff check passed
- pre-push Ruff format check passed
- mypy passed for Linux and Windows platforms
- full pytest suite passed
- documentation contains ordinary inline links instead of unsupported footnote references

Documentation only. No runtime behavior changes.

Related to #106.

### DIFF
--- a/docs/research/EXTERNAL_SWITCH_SCANNING.md
+++ b/docs/research/EXTERNAL_SWITCH_SCANNING.md
@@ -1,0 +1,798 @@
+# Secure external switch scanning for Alpha-OSK
+
+## Executive summary
+
+Switch scanning is a selection method, not a device protocol. A physical switch, keyboard key,
+game-controller input, touchscreen contact, or camera gesture produces a small set of signals. A
+scanner turns those signals into navigation through a changing set of targets, highlights the
+current target or group, and activates the user's choice.
+
+Existing products consistently separate six concerns:
+
+1. acquiring and filtering switch input;
+2. discovering targets;
+3. ordering targets into a scan pattern;
+4. presenting visual and optional auditory feedback;
+5. interpreting selection and recovery commands; and
+6. invoking the chosen action.
+
+The products differ mainly in who owns the targets. An AAC app or on-screen keyboard can scan its
+own cells directly. A system-wide scanner normally discovers semantic controls through the
+operating system's accessibility tree, then falls back to point scanning for inaccessible or
+coordinate-only content. [Apple Switch Control](https://support.apple.com/en-ca/119835),
+[Android Switch Access](https://support.google.com/accessibility/android/answer/6301497),
+[ChromeOS Switch Access](https://support.google.com/chromebook/answer/10274137), and
+[Switchify](https://switchifyapp.com/how-it-works/) demonstrate the system-wide model.
+[Windows On-Screen Keyboard](https://support.microsoft.com/en-gb/windows/use-the-on-screen-keyboard-osk-to-type-ecbb5e08-5b4e-d8c8-f794-81dbf896267a),
+[Grid 3](https://hub.thinksmartbox.com/knowledgebase/using-switch-scanning-with-grid-3/),
+[PRC-Saltillo software](https://documentation.prc-saltillo.com/docs/switch-scanning-setup), and
+[TD Snap](https://us.tobiidynavox.com/pages/td-snap-scanning) demonstrate the app-owned model.
+
+For Alpha-OSK, the scanner-neutral design is therefore:
+
+- expose visible keys and prediction pills as ordinary accessible controls;
+- make each control's accessible action identical to one primary click;
+- give scanners stable ordering metadata and generation-sensitive identity;
+- expose only visible labels and presentation state;
+- preserve the keyboard's no-focus and privacy invariants; and
+- let each scanner own switch capture, timing, scan strategy, highlighting, and auditory output.
+
+On Windows, Microsoft UI Automation (UIA) is the appropriate first adapter. Qt Quick already has
+the semantic properties and press action needed for this provider surface, as documented by the
+[Qt Accessible QML type](https://doc.qt.io/qt-6/qml-qtquick-accessible.html). The current Qt
+Windows bridge exposes those semantics through UIA, including bounds, identifiers, state, and
+Invoke. Because the pinned Qt 6.11 Windows backend does not forward every geometry and structure
+event needed by a live overlay, a scanner should begin with bounded cached polling and add a
+one-way revision notification only if measurements require it.
+
+This interface is scanner-neutral, but it is not client-exclusive. Accessibility APIs authorize a
+class of assistive clients according to each operating system's rules. They do not prove that the
+caller is one named partner. A partner-only requirement is a separate security policy and would
+need a mutually authenticated local channel.
+
+## How switch scanning normally works
+
+### The interaction loop
+
+A typical scan cycle is:
+
+1. **Acquire input.** A switch interface reports press and release, often as a keyboard, mouse,
+   game-controller, Bluetooth, or platform accessibility input. Software-only sources can include
+   screen contact and camera gestures.
+2. **Condition input.** The scanner applies hold-to-accept, release-to-accept, debounce, repeat,
+   and long-press rules before treating the signal as intentional.
+3. **Build a target snapshot.** The scanner uses targets it owns, an accessibility tree, or a
+   geometry-based fallback.
+4. **Choose a scan unit.** The unit may be one item, a row, a column, a block, a semantic group, or
+   a moving line.
+5. **Present the current choice.** A border, fill, magnification, progress indicator, sound, or
+   spoken label tells the user what a switch press would select.
+6. **Advance or select.** A timer advances automatically, or one switch advances while another
+   selects. Selecting a group descends into that group. Selecting an action invokes it or opens an
+   action menu.
+7. **Recover.** Back, previous, cancel, stop, scan-count limits, and inactivity reset prevent a
+   missed target from trapping the user.
+8. **Refresh.** Any UI change produces a new target snapshot. The scanner reconciles the highlight
+   without letting an old position activate a new action.
+
+The switch is therefore only the user's low-bandwidth input. The scanner owns a state machine, and
+the application being scanned supplies accurate target semantics and safe actions.
+
+### Common scan strategies
+
+| Strategy | Typical switch setup | Operation | Best fit | Main cost |
+| --- | --- | --- | --- | --- |
+| One-switch auto scan | Select | A timer advances; pressing selects | One reliable movement | Timing pressure |
+| Two-switch step scan | Next and Select | One press advances; the other selects | Two reliable movements | More activations |
+| Three-switch manual scan | Previous, Next, Select | User moves in either direction | Accurate recovery | More switch positions |
+| Single-switch step and dwell | Next | Press advances; stopping for a dwell selects | One switch without moving timer | Selection delay |
+| Inverse or hold scan | Hold or release | Holding advances; release selects or stops | Sustained activation is easier | Release timing |
+| Row-column scan | Usually Select, optionally Next | Select a row, then an item within it | Regular keyboard or AAC grids | Two-stage errors |
+| Block or group scan | One or more group choices | Repeatedly narrow a set | Large target collections | Group comprehension |
+| Point scan | Select | Stop moving horizontal and vertical lines | UI without usable semantics | Slower and coordinate-sensitive |
+| Radar scan | Select | Stop an angle, then a distance | Sparse or awkward geometry | Precision and timing |
+
+The categories overlap. For example, row-column can advance automatically or manually.
+[Android](https://support.google.com/accessibility/android/answer/6301497) uses row-column
+automatically for keyboards when linear scanning is selected, while
+[Grid 3](https://hub.thinksmartbox.com/knowledgebase/using-switch-scanning-with-grid-3/) and
+[PRC-Saltillo](https://documentation.prc-saltillo.com/docs/switch-scanning-setup) expose row,
+column, block, and cell choices directly.
+
+### Input accommodations are core behavior
+
+Products consistently treat timing and switch conditioning as first-class settings rather than
+implementation details:
+
+- scan interval and a longer pause on the first target;
+- a finite or unlimited number of scan cycles;
+- press versus release activation;
+- minimum hold time before acceptance;
+- ignore-repeat or debounce time;
+- delay before held input repeats;
+- restart location after activation or inactivity;
+- auto-resume versus explicit restart; and
+- long-press actions and reverse direction.
+
+[Android Switch Access](https://support.google.com/accessibility/android/answer/6301497) exposes
+scan time, first-item delay, scan count, repeat suppression, and release-to-act.
+[Apple Switch Control](https://support.apple.com/guide/mac-help/mh43181/mac) exposes
+hold-to-accept, ignore-repeat, repeat delay, restart position, and auto-resume.
+[Grid 3](https://hub.thinksmartbox.com/knowledgebase/using-switch-scanning-with-grid-3/)
+adds press or release acceptance, accidental-press filters, reverse scan, maximum cycles, and a
+switch tester. No single timing value is generally correct. The architecture must allow the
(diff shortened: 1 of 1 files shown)


owenpkent:
## Review

Thorough landscape survey and the threat-model section asks the right questions. The problems are concentrated in the prescriptive parts (the bridge-method contract, the target model, the acceptance tests), where the document describes a dispatch path that does not match how `Main.qml` and `keyboard_bridge.py` actually work. Someone implementing from these sections would build something that fails silently.

### 1. The bridge-method contract omits modifier keys

Line 402 enumerates the only three dispatch routes: "Character keys continue through `KeyboardBridge.pressKey` or `pressKeyLiteral`, special keys through `pressSpecialKey`, and prediction pills through `pressPrediction`."

`qml/Main.qml:3117-3124` dispatches `kd.type === "modifier"` through `toggleShift` / `toggleCapsLock` / `toggleCtrl` / `toggleAlt` / `toggleWin`, none of which appear.

An implementer following line 402 wires Shift's `Accessible.onPressAction` to `pressSpecialKey("shift")`. `src/keyboard_bridge.py:1869` does `_SPECIAL_KEY_NAMES.get(key_name, key_name)`, and that dict (line 1380) has no `"shift"` entry, so `"shift"` falls through raw to the synthesizer. No `hold_modifier()` runs, `_shift_active` never flips, `shiftActive` never updates the keycap. A switch user can never type a capital letter or fire any chord, and nothing reports an error.

Lines 705 and 710 of the same document then require modifier coverage and sticky/locked toggle state, so the contract asks for something its own dispatch list cannot deliver.

### 2. The layer-key acceptance test is unsatisfiable, and satisfying it breaks a documented invariant

Lines 709-710 list "page changes" among the actions that must "reach the same bridge method" as a pointer click.

Layer keys deliberately reach no bridge method. `qml/Main.qml:3125-3161` handles `kd.type === "layer"` as `keyboard.releaseShift(); root.activeLayer = kd.target`, with an in-code comment saying why: "Deliberately does NOT go through keyboard.setLayout: that would persist as the user's layout preference and report the symbol layer from getCurrentLayout()".

An implementer driving this checklist to green adds a `setLayout` call. After a switch user taps `?123`, the compact symbol layer is written to `appSettings` as their chosen keyboard layout and `getCurrentLayout()` reports `qwerty-compact-sym`.

The contract needs a category for "view-state change, no bridge call", or the checklist needs to exempt layer keys.

### 3. One primary action per target leaves right-click-only capabilities unreachable

Line 470 defines "Primary action | One non-retrying activation" and line 434 requires it be "identical to one primary click". But `qml/Main.qml:3176-3200` puts `lockModifier(kd.action)` and the shifted-glyph `pressKeyLiteral` path behind `onKeyRightPressed` only.

So under this contract a switch-scanning user cannot:

- lock Ctrl or Shift, which is the project's documented accessibility answer to "hold the key down"
- type a shifted character without a sticky-Shift round trip
- open a prediction pill's context menu (Show more, Show less, Remove, Edit)
- program an F-key by right-clicking it

Line 469 of the same contract lists "Toggle state | Inactive, sticky, or locked" as a field to expose, promising a state the contract provides no way to enter.

This is the same reachability argument CLAUDE.md makes repeatedly, and in the document's own terms: "a dwell-click, switch-access, head- or eye-tracker pointer, and a single-button adaptive mouse all have no right button" is why the Snippets Manage toggle and the Function Keys settings page exist. The contract needs a secondary action, or the document needs to name which left-click routes cover each right-click capability.

### 4. Edit mode is missing from the action-generation rule

Line 489: "When a prediction, layout key, layer key, or programmed key acquires a different action, Alpha-OSK should replace its accessible object and increment the action generation."

`setEditMode(true)` short-circuits `pressKey` and `pressSpecialKey` for every key at once, emitting `editKeyTyped` / `editSpecialPressed` into the prediction-edit popup or snippets editor instead of synthesizing to the foreground app. Same target, same label, same bounds, completely different action, and by this rule no generation bump.

A scanner holding a reference taken before the snippets editor opened passes the document's own staleness check at 486-487 (still enabled, showing, same ID, same generation, same label) and types into the editor field the user never opened, or the reverse.

### 5. Requiring "identical to primary clicks" pollutes the learned pointer model

`src/keyboard_bridge.py:1662` runs `self._predictor.observe_press(char, offset[0], offset[1])` unconditionally inside the not-privacy branch of `_press_char`, and the overloaded slot defaults `offset` to `(0.0, 0.0)`, meaning key centre, for any one-argument caller. An accessible Invoke is exactly that caller.

`PointerModel` shrinks each slot's mean toward the global mean with `PRIOR = 10`, so a scanning session feeding thousands of exact-zero offsets drags every slot's learned bias toward zero. For the mixed-input user (scans sometimes, uses a mouse other times) the correction the prefix beam relies on is erased by the scanner path, and it is this requirement that causes it.

The contract should say accessible activations are excluded from `observe_press`.

### 6. The discovery procedure assumes a single top-level window

Line 546: "A scanner should locate Alpha's top-level UIA window, verify the owning process and product, then query descendants with the contract identifier prefix."

Alpha-OSK presents several. `snippetsWindow` and `symbolsWindow` are separate top-level `Window` instances found by `objectName` in `keyboard_app.py::_wire_floating_windows`, plus the settings window, Help, and `vizWindow`, and `KeyActionEditor` and `predEditPopup` are parented to `Overlay.overlay`.

A scanner implementing this finds only the keyboard grid. The switch user can type letters but cannot reach Snippets, Symbols and Emoji, Settings, the Dashboard, or the key-action editor, which is every non-typing feature in the product.

### 7. The Windows authorization conclusion reasons from the wrong install location

Line 586 reasons from "Alpha's standard-user installation" to conclude the practical contract is integrity-level based.

`build/windows/build.py:619` defines `!define INSTALL_DIR "$PROGRAMFILES64\Alpha-OSK"` and `run.py:271` `ensure_admin_windows()` re-launches under UAC. Lines 655-657 of this same document say Alpha-OSK publishes "configurations for signed UIAccess applications installed under Program Files", and 583-584 cite the Microsoft policy that UIAccess requires a protected install location. So the document contradicts itself, and a reader sizing the threat model from 586 assumes the wrong integrity level throughout the "which scanners can reach this UI" analysis.

### 8. The UIAccess escalation is not named

Line 592 concedes "another qualifying assistive client may be able to inspect and invoke Alpha's exposed controls" and 594 calls this "suitable when interoperability with assistive technologies is the policy goal".

`build/windows/alpha-osk.exe.manifest:93-95` sets `requestedExecutionLevel level="asInvoker" uiAccess="true"`, and the manifest's own comment says this lets the keyboard "send keystrokes to ELEVATED windows on the regular desktop (e.g. an admin Command Prompt, Task Manager, Registry Editor)".

So a UIA client that invokes Alpha's keys gains synthesized keystroke delivery into high-integrity windows it could not reach itself, with Alpha-OSK's EV-signed UIAccess token doing the crossing. That is the one property distinguishing Alpha from an ordinary invokable app, and the "suitable" verdict is reached without weighing it.

### 9. Smaller items

- **Bounds has no owning item specified.** Line 466 requires a global screen rectangle and 731 requires the external highlight stay aligned, but nothing says whether the `Accessible` attached properties go on the `KeyButton` root or its `MouseArea`. The MouseArea is anchored with negative `hitMarginH`/`hitMarginV` so it reaches half a gap past the keycap on every side, and vertical neighbours deliberately overlap. Qt derives `QAccessibleInterface::rect()` from the item carrying the node, so the MouseArea choice hands a point-scanning client overlapping rectangles and an overlay wider than the key it names. Worth stating the node belongs to the keycap item.
- **Source citations use moving refs.** Line 562 cites `qt/qtbase/blob/6.11/...` for the precise claim that "LocationChanged, ObjectShow, and ObjectHide are not forwarded"; 667 and 669 cite `switchify-pc/blob/main/...`. Both are branches. When Qt lands a 6.11.x patch the link still resolves and now contradicts the sentence citing it. `requirements.txt` pins `PySide6==6.11.1`, so the v6.11.1 tag or a commit SHA is the right anchor.
- **No inspection date.** Lines 654-658 assert the inspected Alpha-OSK 1.4.1 and Switchify 1.0.0-rc.6 installers had valid signatures and matching digests, 648 says "its current public product description", 664 says the manifest "does not declare a DPI mode". Nothing says when. Switchify ships rc.7 and Alpha-OSK ships 1.4.2, and the Evidence boundaries section asks the reader to recheck. Every measured claim elsewhere in this project is dated ("measured 2026-09-02").
- **Mind Express 5 is cited through two different URLs.** Line 192 uses `/en/mind-express/`, lines 153 and 770 use `/en/mind-express-5/`, for the same product and the same claims.
- **tecla-e appears twice in the reference index as if distinct.** Lines 783-784 list `/products/tecla-e` and `/pages/tecla-e`. Both inline citations (283, 343) use only `/products/`, so the `/pages/` entry supports no claim.
- **Blue2 FT is in two contradictory rows.** Line 324 lists it under "Proximity switch | Movement near a sensor, with little or no pressure"; line 326 lists the same product under "Integrated wireless switch". It is a two-surface Bluetooth switch that is pressed. A clinician using the proximity row to shortlist for a user who cannot tolerate contact pressure orders the wrong thing.
- **Not in `docs/README.md`.** Lines 59-73 list every other file under `research/` with a one-line description. Nothing anywhere in the repo references this file, so a contributor working from CLAUDE.md or the docs index never learns the contract exists. Worth deciding placement at the same time: the Delivery sequence and Acceptance tests sections read as a plan to implement, which `docs/README.md` puts in `roadmap/` rather than `research/`.

docs/research/EXTERNAL_SWITCH_SCANNING.md
```diff
@@ -0,0 +1,798 @@
...
+modifiers. `Accessible.focusable` must be set consistently with the keyboard's no-focus design,
+rather than accepting the Button role's default.
+
+### Discovery, caching, and coordinates
+
+A scanner should locate Alpha's top-level UIA window, verify the owning process and product, then
+query descendants with the contract identifier prefix. Microsoft documents that
+[UIA caching](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-cachingforclients)
+can retrieve several properties for several elements in one cross-process call. Cached patterns
+cannot perform actions, so the client must retain or reacquire a live element for the final Invoke.
```
owenpkent: [P3] Qualify this restriction with `AutomationElementMode_None`

The statement that cached patterns cannot perform actions is too broad. In the cited Microsoft page, that restriction is specifically under the cache-only `AutomationElementMode_None` case. The default full-reference mode retains access to the underlying UI; caching an Invoke pattern does not itself make that pattern non-actionable.

Specify either a cache request with `AutomationElementMode_Full` and a retained Invoke pattern, or a cache-only snapshot followed by reacquisition before activation. This makes the polling design implementable without an unnecessary lookup and keeps cache freshness distinct from whether an action can be invoked. See [Microsoft's cache reference-strength section](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-cachingforclients#strength-of-element-references) and [AutomationElementMode](https://learn.microsoft.com/en-us/windows/win32/api/uiautomationclient/ne-uiautomationclient-automationelementmode).

owenpkent (commented):
## Follow-up review of `e3bb83d`

There is one additional API clarification below, plus a correction to the earlier review and a concrete CI blocker.

### Correct the UIA caching guidance

The inline comment identifies the distinction between cached data and a full element reference. State the required `AutomationElementMode` explicitly so an implementer can retain cached properties and an invokable pattern without adding an unnecessary rediscovery step.

### Correction to the earlier hardware comment

The earlier review says Blue2 FT is a pressure-operated switch and therefore cannot appear in the proximity row. AbleNet's [Blue2 FT product specification](https://www.ablenetinc.com/blue2-ft/) explicitly describes proximity sensing, activated by being near or touching the surface. It also includes Bluetooth and external switch jacks. The two rows describe different dimensions, activation mechanism and transport, so their overlap is valid. Please retain the proximity classification.

### Reconcile the research proposal with #114

The target grammar here is a proposal. #114 now supplies a more specific measured contract, including `aosk.v1.*` identifiers, speakable fallback names, the UIA `FullDescription` mapping, generation-sensitive prediction delegates, and window-state handling. Add a status note and cross-reference when that contract lands so readers do not implement these two identifier schemes as competing current interfaces. The earlier dispatch feedback can be resolved by specifying the same QML action handler, including modifier toggles and layer-local state changes, rather than an exhaustive list of bridge methods.

### CI and validation

[CI run 34555570837](https://github.com/owenpkent/alpha-osk/actions/runs/34555570837) failed in OSV. Its logs identify development dependencies `js-yaml` 4.3.1 and `sharp` 0.35.3 in `backend/cf-worker/package-lock.json`, with fixes reported as 4.3.2 and 0.35.4. The eight test shards, both coverage jobs, lint, and type checks passed. These dependencies are inherited by this documentation-only PR; bring in the dependency fix and rerun CI after resolving the currently reported merge conflict.

Validation for this review: read the proposed contract, traced relevant QML dispatch, checked the local manifest link, and spot-checked the cited Microsoft, Qt, and AbleNet documentation. Installer signatures, hardware behavior, and packaged cross-platform scanning were not remeasured.

owenpkent:
## Addressed in 9345e36

Rebased onto current `main`, which carries the `js-yaml` 4.3.2 / `sharp` 0.35.4 fix, so the OSV failure is cleared without this documentation PR carrying a dependency commit.

- **Dispatch contract (1, 2, 3):** resolved the way the follow-up review suggested, by naming the QML action handler rather than enumerating bridge methods. An accessible activation is defined as raising `KeyButton.keyPressed`, and reaching whatever that key's `kd.type` handler does. Modifier keys are named. **Layer keys are a fourth category that reaches no bridge method**, so the conformance test exempts them and asserts the view-state change instead. For the four right-click-only capabilities, no second UIA action is invented: programming an F-key, capitals and shifted symbols get their left-click routes named, and the prediction context menu and the right-click lock are recorded as known gaps rather than papered over, so the "locked" state field is described as observable but not enterable.
- **Reconcile with #114:** status note after the H1 saying the contract, UIA-design, delivery and acceptance sections are the proposal and that `docs/architecture/UIA_TARGETS.md` is what to implement against, plus inline cross-references to its sections (Finding the keyboard, Identity, State, Bounds and DPI, The revision beacon, Stale predictions cannot fire, What this does and does not protect, Independent verification). The two identifier schemes are no longer presented as competing current interfaces.
- **Edit mode (4):** added to the action-generation rule, with a note that UIA_TARGETS.md does not yet cover it.
- **Pointer model (5):** stated as a contract requirement, with the honest note that #114's `activateFromAssistiveClient()` does not enforce it at its current head. **I am carrying that over as a change to #114 itself**, since a scanning session feeding exact-centre presses erases the learned bias.
- **Discovery (6):** rewritten for the six top-level windows and popups, citing #114's keyboard-only v1 scope.
- **Windows authorization (7, 8):** reasoned from the Program Files install and the `uiAccess="true"` manifest rather than a standard-user install, and the escalation is named: a UIA client that invokes Alpha's keys gains synthesized keystrokes into high-integrity windows it could not reach itself, with Alpha's UIAccess token doing the crossing.
- **Smaller items:** the accessible node belongs to the `KeyButton` root rather than its oversized `MouseArea`; qtbase links pinned to `v6.11.1` and switchify-pc links to commit `7e383f4` (all verified 200); "as inspected on 2026-09-11" with an rc.7 / 1.4.2 recheck note; Mind Express unified on one URL; the unused `/pages/tecla-e` entry dropped; a `docs/README.md` entry added under `research/`.
- **Caching (inline P3):** rewritten to distinguish `AutomationElementMode_Full` with a retained Invoke pattern from a cache-only `None` snapshot needing reacquisition, citing both Microsoft pages.
- **Blue2 FT:** left in the proximity row, per the follow-up review's retraction.

PR #112, opened 2026-09-11
fix(cf-worker): bump sharp and js-yaml to clear two new advisories

owenpkent:
## Summary

- Bumps `sharp` to 0.35.4 and `js-yaml` to 4.3.2 in the telemetry worker's `overrides` block, clearing two high-severity advisories published on 2026-09-08.
- Both are dev-only transitive dependencies of Wrangler. Neither ships in the keyboard or runs on the deployed worker; what they block is development, because the OSV gate is `fail-on-vuln: true` by design.
- Syncs the override list in `docs/build/RELEASE.md`, which had drifted from the file it documents.

## Related issue

Closes the Dependabot alert for GHSA-rgj7-g3m4-5g8c (alert #37).

`js-yaml` GHSA-2883-xcg3-v3hh was not raised as a Dependabot alert but is
flagged by both `npm audit` and `osv-scanner`, so it would have failed the
same CI gate. Fixed here rather than left for a second PR.

## What the two advisories actually are

**GHSA-rgj7-g3m4-5g8c** (sharp, CVSS 8.9) is not a flaw in sharp. sharp
bundles libvips, libvips bundles libheif, and two libheif advisories
(GHSA-g89c-p67h-r497, GHSA-2jg2-4ch7-h545) reach the lockfile through that
chain. The fix is the libvips 1.3.3 rebuild that 0.35.4 pulls in, which is
why the diff is 27 `@img/sharp-*` platform packages rather than one. It
reaches us via `wrangler -> miniflare -> sharp`.

**GHSA-2883-xcg3-v3hh** (js-yaml, CVSS 7.5) is a real parser flaw:
`maxTotalMergeKeys` does not bound CPU use for empty merge sources. It
reaches us via the CycloneDX SBOM generator (`xmlbuilder2 -> js-yaml`,
which wants `^4.1.1`, so staying on the 4.x line keeps it compatible).

Both packages were already covered by the `overrides` block and only needed
their floors raised, so this is that block working as intended rather than a
gap in it.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs only
- [x] Build / CI / tooling
- [ ] Other (describe)

## Test plan

- `osv-scanner --lockfile=requirements-dev.txt --lockfile=backend/cf-worker/package-lock.json`, the same two lockfiles the CI job scans: **0 vulnerabilities**.
- The same scan against `origin/main`'s lockfile, as the control: **2 packages, 2 high**, confirming this is a live CI failure on main and not only a Dependabot notice.
- `npm audit` in `backend/cf-worker`: 0 vulnerabilities (it reported 1 high before the js-yaml bump).
- Diffed every version change in the lockfile against main: 29 packages, all of them sharp's platform binaries plus their `@emnapi/runtime`, and js-yaml. No movement in `wrangler`, `typescript` or `@cloudflare/workers-types`.

- [x] `python check.py` passes locally (ran via the pre-push hook: ruff, both mypy platforms, pytest, 110.7s)
- [ ] Added or updated tests
- [ ] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

No tests added: nothing in the suite reads these files, and the behaviour
under test is the CI gate itself, which this PR's own osv-scan run exercises.
No `run.py` check for the same reason, since no code the keyboard loads is
touched.

## Accessibility check

n/a. No runtime code changed.

## Notes for reviewers

Per CLAUDE.md, this touches the telemetry worker, so to be explicit about the
part that needs alignment: **the telemetry payload and the consent model are
unchanged.** This is only the dev-dependency graph of the worker's local
tooling. `src/worker.ts`, `schema.sql` and `DEFAULT_ENDPOINT` are untouched.

The `docs/build/RELEASE.md` edit is worth a glance because it is the kind of
drift that recurs: that paragraph listed five versions that had since moved
and omitted `ip-address` and `sharp` entirely, having gone stale at the 1.2.x
override refresh. It now also says that `package.json` is the authority when
the two disagree, so the next reader does not have to work out which one to
trust.

One thing deliberately not done: `js-yaml` 5.x is out, but the override is a
floor and `xmlbuilder2` asks for `^4.1.1`, so 4.3.2 is the minimal fix.
Likewise `sharp` is left as a `^0.35.4` floor rather than an exact pin, which
matches every other entry in the block.


owenpkent:
## Review

**The fix itself is correct and complete.** Verified independently: queried the OSV API for every advisory cited, batch-scanned all 266 lockfile packages on both `main` and this branch, and checked all 30 changed `integrity` hashes against the npm registry.

- `main` has exactly 2 vulnerable packages: js-yaml 4.3.1 (GHSA-2883-xcg3-v3hh) and sharp 0.35.3 (GHSA-rgj7-g3m4-5g8c). This branch has 0.
- Fixed versions match the advisories exactly (js-yaml 4.3.2, sharp 0.35.4).
- Exactly 29 package entries moved: all sharp platform binaries plus `@emnapi/runtime` plus js-yaml. No drift in wrangler or typescript.
- Every integrity hash matches the registry, and every other GHSA cited in the rewritten docs checks out.

Everything below is documentation accuracy and a root-cause gap. Nothing blocks the merge.

### 1. The CHANGELOG attributes js-yaml to Wrangler, and contradicts itself in the same sentence

Line 10 reads: "Both arrive as dev-only transitive dependencies of Wrangler, the local development tool: sharp through miniflare, js-yaml through the SBOM generator."

Traced against the lockfile, js-yaml is required only by `xmlbuilder2`, required by `@cyclonedx/cyclonedx-npm`, which is its own root devDependency (`"@cyclonedx/cyclonedx-npm": "^5.0.0"`). Wrangler appears nowhere on that path. Only sharp goes wrangler to miniflare to sharp.

This matters beyond pedantry because the CHANGELOG is the permanent record of a security fix, and the reader's next question ("can this dependency be dropped or bumped away?") has a different answer for the SBOM generator, which runs in the `predeploy` chain before every `wrangler deploy`, than for Wrangler itself.

The PR body's detail section gets it right ("It reaches us via the CycloneDX SBOM generator (xmlbuilder2 to js-yaml)"), so only the summary line is wrong. Note the same claim is already at CHANGELOG.md:277 from the previous override refresh, so this is the second time it has been written down.

### 2. `ip-address ^10.3.1` is the only override in the list with no advisory ID

Every other entry in that sentence names the GHSA that justifies it (`tar ^7.5.21` to GHSA-23hp-3jrh-7fpw, `fast-uri ^3.1.5` to GHSA-4c8g-83qw-93j6, and so on). `ip-address ^10.3.1` is bare.

The advisory is GHSA-mwp4-54f8-5fhr: Address4 decodes leading-zero octets as decimal while resolvers decode them as octal, allowing SSRF and trust-boundary bypass. Fixed at exactly 10.3.1, which is why the floor is that precise number.

The whole point of this paragraph, which this PR restates, is to be the record of why each pin exists. An entry with no ID cannot be evaluated for retirement later, so the next maintainer either keeps a pin forever or removes one that is still load-bearing. The gap is inherited from CHANGELOG.md:277, which also added the pin without naming the advisory.

### 3. The doc-sync pass missed the stale osv-scanner pin one paragraph up

`docs/build/RELEASE.md:25` says the job is "pinned to `google/osv-scanner-action@9a498708959aeaef5ef730655706c5a1df1edbc2` (v2.3.8)".

`.github/workflows/ci.yml:337` actually pins `google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@6e4298ebc4db23e847df9b2e2de2939d6f066c67 # v2.5.1`. Different SHA, different version, and a different reference form (reusable-workflow path rather than action root). `git log -S` dates the change to PR #44.

This PR's third stated goal is to sync the override list in `docs/build/RELEASE.md` "which had drifted from the file it documents", and this is the same drift class against the same file, two paragraphs apart in the same section. A reader following line 27's own instruction to "Bump the `google/osv-scanner-action` pinned SHA" would diff against the wrong baseline.

The same stale SHA and version also sit in `docs/build/WINDOWS.md:690`, `docs/WHITEPAPER.md:561`, and `docs/paper/alpha-osk.tex`.

### 4. Root cause: dependabot has no npm entry for the worker

`.github/dependabot.yml` has exactly two `updates` entries, both at `/`: `pip` and `github-actions`. There is no `npm` ecosystem for `/backend/cf-worker`, so Dependabot never opens a version-update PR against that lockfile. And Dependabot *security* updates cannot fix a transitive held by an npm `overrides` block anyway, since it does not resolve overrides.

So the only signal left is `osv-scan` turning red, which by design happens on an unrelated PR that then cannot merge. That is exactly what the PR body describes ("what they block is development") and what this PR exists to unblock. The PR itself notes js-yaml GHSA-2883-xcg3-v3hh "was not raised as a Dependabot alert", so half of this change had no automated signal at all.

A three-line `npm` entry at `directory: "/backend/cf-worker"` turns the next occurrence into a bot PR instead of a blocked branch plus a hand-written fix.

### 5. The doc-drift fix is a re-sync, not a mechanism, and this paragraph has now gone stale twice

The remedy applied is to retype the eight versions by hand and add "`package.json` is the authority if this paragraph disagrees with it". That sentence tells a reader which copy to distrust; it does nothing to stop them diverging, and the next override bump (which by the paragraph's own admission "moves whenever a new advisory lands rather than on a release") silently restales it.

This repo already has the right pattern: `tests/test_windows_installer.py` parses the generated NSIS script and expands the `!define`s before comparing, so source and doc cannot drift unnoticed. The equivalent here is a test that loads `backend/cf-worker/package.json`'s `overrides` and asserts every name and range appears verbatim in `docs/build/RELEASE.md`. Cheaper than the two hand-syncs already spent, and it fails loudly instead of being noticed on a third pass.

The alternative deep fix is to drop the version numbers from the prose entirely and keep only the advisory IDs and reasoning, which do not drift.

### 6. Smaller items

- **The sharp entry names libvips but not libheif.** The new text says the advisory "is cleared by the libvips 1.3.3 rebuild 0.35.4 pulls in rather than by anything in sharp itself". The advisory's own OSV summary is "sharp: Vulnerabilities in libheif: GHSA-g89c-p67h-r497 and GHSA-2jg2-4ch7-h545". The flaws are in libheif, which libvips bundles; libvips 1.3.3 is the vehicle rather than the subject. CHANGELOG.md:8 gets the full chain right, but RELEASE.md is the doc someone opens when the next sharp advisory lands, and from it alone they would search for a libvips advisory and not find one.
- **The sharp override is a breaking bump, not a floor.** miniflare declares `"sharp": "0.34.5"` exactly; the override rewrites it to `^0.35.4`, crossing a 0.x minor boundary. sharp 0.35 changed the prebuilt-binary layout (the `@img/sharp-libvips-*` split visible in this diff) and raised `engines.node` to `>=20.9.0`. Nothing verifies the result: `ci.yml` has no `npm ci` or `npm install` job at all, and the test plan runs osv-scanner and `npm audit` but never `wrangler dev --local`. Practical impact is low, since `wrangler.toml.example` declares only D1 and rate-limit bindings and no Images binding so miniflare never loads sharp, and the override already crossed this boundary at `^0.35.0` on main. Worth knowing only because both the doc and the PR body call these entries "floors", which understates what this one does.
- **One em dash survived in a line you rewrote.** `docs/build/RELEASE.md` line 27 contains "so `npm run dev` / `npm run deploy` need Node 22 — but `npm install`, the SBOM, and the CI osv-scan all run fine on Node 20". It predates this PR, but the whole line was rewritten so the diff presents it as added. A comma or semicolon reads identically. The CHANGELOG additions are clean and the commit message carries no em dash.

PR #113, opened 2026-09-11
fix(prediction): let a skipped apostrophe still find the word

owenpkent:
Typing `ill` offered `illinois`, `illegal` and `ill` but never `I'll`. `hes` never offered `he's`, `im` never offered `I'm`, `shes` never offered `she's`.

## Why it happened

The apostrophe costs an extra click on this keyboard, and a layer hop on the compact layouts, so it is the character a user skips deliberately. That is not a mis-click, and the fuzzy source structurally could not rescue it:

- The prefix beam reaches `i'll` from `ill` only through an omitted click at `LOG_OMIT` (-2.5), which is past `FREQUENCY_MAY_BUY` (-1.5), so `_protect_exact_completions` clamps it below every completion of the live prefix `ill`. There are eight.
- `he's` was not reached at all.
- `im` is two characters, so the beam's `MIN_TYPED` guard refuses it outright.

The contractions that did work (`dont`, `cant`, `its`, `whats`) worked only because their exact-completion families happen to be small enough that the demoted candidate still surfaced at rank 4 or 5.

## The fix

One clause in `NgramPredictor._matches_partial`, the single choke point every prefix match goes through: a word containing an apostrophe also matches a typed prefix its apostrophe-stripped form starts with, **and only while the user has typed no apostrophe themselves** (once they have, `don'` already matches `don't` exactly, and stripping on top would make the prefix mean less than what was typed rather than more).

Putting it in the exact source rather than the fuzzy one is the point: the contraction then competes on its own count instead of having to buy its way past eight exact completions. It is the same reasoning `_APOSTROPHE_INSERTION_PROB` (0.50 against a generic 0.15) already encodes for the whole-word path, one layer over. Cheapening the beam's apostrophe omission was considered and rejected: it prices a deliberate skip as a motor error and still leaves the clamp to fight.

### Second half: typing the apostrophe blanked the bar

`_press_char`'s gate on whether to re-query was `char.isalpha()`, so the apostrophe threw the bar away at the moment it was right. Typing `don` put `don't` at the top of the bar and the `'` cleared it, one click short of the word. `i'` discarded `I'm` / `I'll` / `I'd` / `I've`, which are the entire reason to type an apostrophe in that position.

That is the same oversight the digit had, and which the comment at that call site already describes. `_continues_a_word` is the gate now: letters always, plus `'` and `_` when there are letters in front of them. A leading `'` still asks nothing, since it carries no prefix and the round trip would return nothing.

## Measurements

**The gain.** On the held-out AAC corpora, taking every contraction occurrence and typing it the way a user of this keyboard does (no apostrophe), the intended word is offered somewhere while typing it:

| | offered |
|---|---|
| before | 71/109 = **65.1%** |
| after | 103/109 = **94.5%** |

`i'm` is the most common contraction in the set (24 occurrences) and was previously unreachable at every prefix length. The 6 remaining misses are possessives (`doctor's`, `today's`, `tracy's`) that are not in the vocabulary as words, which no prefix rule can reach.

**Keystroke savings: unchanged to the decimal** (49.1% aac-dev, 50.4% aac-test, matching the documented baselines). Those corpora type contractions *with* their apostrophes, so the benchmark cannot see this change at all. It is a regression check here, not a measurement of the fix, and the PR is explicit about that rather than quoting a flat number as if it were evidence either way.

**Blast radius.** Sweeping all 4,056 two- and three-letter prefixes, before against after:

- 10 prefixes change (0.2%)
- 7 change by **gaining** a contraction
- **0** change by losing one
- the 9 words displaced anywhere are all rank 5-6 tail items (`impact`, `illustration`, `ideal`, `website`, `yup`, ...), each displaced by a contraction far more common in conversational text

Three prefixes (`fav`, `hen`, `yu`) shuffle at ranks 5-6 without any apostrophe word entering the bar; that is the merge's sum-to-1 normalisation denominator shifting when one more candidate enters the n-gram source, breaking near-ties differently.

## Tests

Every positive is paired with the near-miss it must still reject, and each new test was verified to **fail without the fix**.

- `test_ngram_predictor.py::TestTheApostropheIsOptionalInATypedPrefix` — the predicate. The end-to-end case seeds the contractions explicitly and says why: a bare `NgramPredictor` carries no apostrophe words at all (they arrive with `base_dictionary.txt`, which only the hybrid loads), so an unseeded version would pass against a predicate that had stopped working.
- `test_hybrid_predictor.py::TestASkippedApostropheStillFindsTheWord` — the ranking, on the shipped word lists rather than a stub, because the claim is about real frequencies. Includes the inverse that `ill` keeps offering `ill` and `illinois`, and that a prefix with no contraction behind it grows none.
- `test_keyboard_bridge.py::TestTypingTheApostropheKeepsTheBar` — the gate, with the inverses that a bare `'` still asks nothing and that ordinary punctuation still clears the bar.

Full gate green: `ruff check`, `ruff format --check`, `mypy` under both `--platform linux` and `--platform win32`, and 2546 passed / 4 skipped.

## Flagging per CLAUDE.md

This changes the prediction engine. Two judgement calls worth a look:

1. **`id` now offers `I'd` at rank 1**, ahead of `idea` (which moves to rank 2). Same for `iv` → `I've` ahead of `ivory`, and `wer` → `we're` ahead of `were`. For conversational AAC text I think that is right, but it is a rank-1 takeover on short prefixes and is the one thing here that is a preference rather than a bug fix.
2. Branched from `origin/main`, so this does not include PR #108's short-word work. The two are adjacent (#108 touches `allow_short_prefix` in the same merge) but independent; no conflict expected.

### DIFF
--- a/src/keyboard_bridge.py
+++ b/src/keyboard_bridge.py
@@ -1805,8 +1805,9 @@ class KeyboardBridge(QObject):
             # `_current_word` of "gm" that it reset at the "@" anyway.
             #
             # The old gate here was `if char.isalpha()`, which is why a
-            # digit used to blank the bar outright.
-            if char.isalpha() or self._in_token_context():
+            # digit used to blank the bar outright, and why the apostrophe
+            # of a contraction did too until `_continues_a_word` took over.
+            if self._continues_a_word(char) or self._in_token_context():
                 self._refresh_prediction_bar()
             else:
                 self._clear_token_pills()
@@ -3358,6 +3359,30 @@ class KeyboardBridge(QObject):
     # ------------------------------------------------------------------
     #  Structured-token predictions (numbers, phone numbers, emails)
     # ------------------------------------------------------------------
+
+    def _continues_a_word(self, char: str) -> bool:
+        """Does this keystroke leave a run the word engine can complete?
+
+        Letters always do.  The apostrophe and the underscore are word
+        characters too, the same set `_press_char` keeps in
+        `_current_word` so that "don't" and "snake_case" stay single
+        tokens, but only *inside* a word: a leading one carries no prefix
+        of its own, so asking the engine about it buys nothing and costs a
+        round trip on the keystroke path.
+
+        The apostrophe is the half that was missing.  Typing "don" offers
+        "don't" at the top of the bar, and the `'` used to blank the bar
+        outright, one click short of the word; "i'" threw away the four
+        pills ("I'm", "I'll", "I'd", "I've") that are the whole reason
+        to type an apostrophe there.  That is the same oversight the digit
+        had, one character class over, and it is deliberately a separate
+        question from the word-character rule in `_press_char`: that one
+        decides what `_current_word` keeps, this one decides whether the
+        run so far is worth asking about.
+        """
+        if char.isalpha():
+            return True
+        return char in ("'", "_") and any(c.isalpha() for c in self._current_word)
 
     def _in_token_context(self) -> bool:
         """Is the run before the cursor a structured token in progress?
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,6 +155,66 @@ The last of the 2026-09-02 recommendations, landed 2026-09-03, and the one whose
 
 **The gain is modest, and that is the finding.** `scripts/bench/ksr.py --pointer BIAS_X,BIAS_Y,NOISE` simulates a pointer end to end (reported key, offset, and the presses the bias is learned from) and reports each condition three ways. A pointer whose misses are mostly random (bias 0.2, 0.15; noise 0.3; about a quarter of clicks on the wrong key), the robustness check: keys only 50.5%, plus offsets 50.8%, plus learned bias 50.9%. One whose misses are mostly systematic (0.35, 0.25; 0.15; 22% wrong), the case this feature is for: 51.1%, 51.5%, **51.8%**. The earlier per-key simulation (75% to 86.5% intended-key recovery with a learned bias) was real but did not translate, because the prefix beam plus the dictionary already recover most single-key errors from the reported key alone; what a click position adds on top is a few tenths of a point, more for a systematic pointer than a scattered one. It ships because it never regresses at 0.55, costs nothing at runtime, is privacy-gated like every other learning, and because the persisted table is the first measurement of this user's actual pointer bias, which no simulation can supply. Tests: `tests/test_pointer_model.py`, `tests/test_click_position.py` (recognizer, bridge and persistence hops, each paired with the near-miss it must leave alone) and `tests/test_qml_click_position.py`, which loads `KeyButton.qml` on its own in a `QQuickView` (one item, not a Repeater delegate, so the scene point is reliable under the offscreen plugin) and presses at a known point, and which also calls the overloaded `pressKey` / `pressKeyLiteral` slots with three arguments and with one from a QML function against a real bridge: the Python tests call the slots directly and never touch Qt's overload resolution, and a three-argument call that failed to bind from QML would degrade every press to the key centre with no warning anyone would see.
 
+## The apostrophe is optional in a typed prefix
+
+Typing `ill` offers `I'll`, `hes` offers `he's`, `im` offers `I'm`. The rule is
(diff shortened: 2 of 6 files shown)


owenpkent (commented):
## Review of `155d9e4`

One prediction/typing-state mismatch needs attention: the new continuation gate also enables predictions immediately after an underscore, although the word tokenizer does not preserve that character. The inline comment includes the observed insertion result.

### Apostrophe behavior

The apostrophe-stripped matching clause is appropriately conditional on the typed prefix containing no apostrophe. Once an apostrophe has been typed, ordinary prefix matching retains its meaning. Keeping this in the n-gram candidate source allows contractions to compete on frequency instead of paying a generic omitted-click penalty.

The short-prefix and ranking tests cover the important examples and near misses. **31 focused tests passed** across the new n-gram, hybrid, and bridge cases. The CI run associated with this head also passed.

### Insertion coverage to add

The tests should also tap one of the newly offered contractions. An isolated mocked bridge probe for `ill` -> `I'll` reached `replace_text(3, "I'll ")`, because the apostrophe and capitalization prevent suffix-only insertion. Please cover that ordinary replacement path, the compatibility BackSpace/retype path, held modifiers, and privacy suppression of learning.

The generic replacement path already has documented terminal limitations. This review did not test a live terminal, so that is an existing integration limitation to validate before promising the new contraction behavior in terminal applications. All probes recorded synthesis calls without sending desktop input.

The neighboring short-prefix/personal-learning work is #108; #109 is its stacked architecture-documentation PR. The final paragraph of this PR description currently points to #109 for that implementation.

src/keyboard_bridge.py
```diff
@@ -3359,6 +3360,30 @@ def _update_predictions(self) -> None:
...
+        pills ("I'm", "I'll", "I'd", "I've") that are the whole reason
+        to type an apostrophe there.  That is the same oversight the digit
+        had, one character class over, and it is deliberately a separate
+        question from the word-character rule in `_press_char`: that one
+        decides what `_current_word` keeps, this one decides whether the
+        run so far is worth asking about.
+        """
+        if char.isalpha():
+            return True
+        return char in ("'", "_") and any(c.isalpha() for c in self._current_word)
```
owenpkent: [P2] Keep underscore handling consistent with the prediction tokenizer

`_continues_a_word` includes `_`, but the English word tokenizer keeps letters and apostrophes only. With this newly enabled query, typing `snake_` leaves `_current_word == "snake_"` while the word model predicts from `snake`. An isolated bridge probe on this head displayed `snake` first; selecting it called `replace_text(6, "snake ")`, removing the underscore. Before this change, the underscore cleared the prediction bar.

The n-gram candidate is being treated as an exact completion of a prefix produced by discarding a typed character. Limit the new query condition to the intended apostrophe case until underscore tokenization is supported, or align the tokenizer and bridge word-boundary rules. Add a bridge regression that types the underscore and selects an emitted pill, so prediction display and insertion are checked together.

owenpkent:
## Addressed in 00df57c

- **Underscore (P2):** `_continues_a_word` is now the apostrophe alone. The underscore stays in `_current_word` (that rule is unchanged) but clears the bar as it did before this PR, until the tokenizer and the gate agree on it. The docstring and CLAUDE.md say why. Regression: `test_an_underscore_still_clears_the_bar`, which types `snake_` and asserts an empty bar; it fails against the previous gate (the bar held `snake`) and passes now.
- **Contraction tap coverage:** new `TestATappedContractionInsertsWhatItDisplayed` types `ill`, asserts `I'll` is offered, and taps it through the ordinary path (`replace_text(3, "I'll ")`), Compatibility Mode (three BackSpaces then `I'll `), a locked Ctrl (released before the replace, re-held after, lock intact) and privacy mode (still inserts, learns nothing).
- **Description:** the short-word cross-reference now points at #108, not #109.
- Rebased onto current `main`.

Local gate: ruff, format, mypy (both platforms) pass. The full pytest run hit the two order-dependent `TestATwoLetterMisClickDoesNotEmptyTheBar` cases under xdist; they pass alone and are untouched by this PR. #114's 70ac8d1 builds the beam before those tests read it.

PR #114, opened 2026-09-12
feat: expose keys and predictions as UI Automation scan targets

owenpkent:
Closes the Alpha-OSK half of #106. Every visible key and prediction pill is a UI Automation `Button` that an external switch-scanning application can find, highlight and press. A scanner can also minimize the keyboard, restore it without taking focus, and read which state it is in.

The normative contract is `docs/architecture/UIA_TARGETS.md`. It records what each decision rests on and how each part was verified.

## Why UI Automation rather than the IPC server that was asked for

The deciding argument is authorization. Alpha-OSK ships with `uiAccess="true"`, so it can send input to windows an ordinary same-user process cannot. A pipe that validated only the user would lend that ability to anything running as the user, which makes the keyboard a confused deputy. Over UIA, Windows decides who may drive the keyboard, and we write no ACLs, framing, PID or token checks, or publisher policy. The reporter confirmed that the audience is any Windows-permitted assistive technology, not Switchify alone, which is what UIA models.

## The contract

- **Finding the window.** AutomationId `alphaOsk.alphaOskKeyboard`.
- **Targets.** Ids are `aosk.v1.<section>.<row>.<index>` across `grid`, `fn1`, `fn2`, `num`, `nav` and `pad`, and pills are `aosk.v1.pred.<index>.g<generation>`. The format is defined once on `Main.qml`'s root and called by all six surfaces that draw targets.
- **Presence means activatable.** Hidden panels, unshown layers and a minimized keyboard contribute no targets.
- **Revision beacon.** One property folds in geometry, layout, panels, layer, every modifier and lock, the prediction generation and the window's visibility. A scanner polls it and re-snapshots only when it moves.
- **Invoke.** It does what a primary click does, never arms the repeat timer, and never takes focus.
- **Window state.** Clients use the window's standard `WindowPattern`: `SetWindowVisualState(Minimized)`, `SetWindowVisualState(Normal)` and `WindowVisualState`. There is no Alpha-OSK-specific control for this.

## What measurement changed

Each of these would have shipped wrong on the strength of the documentation or of a first measurement.

- **Lock state rides in `FullDescription` (30159), not `HelpText`.** Qt maps `Accessible.description` there. It is UIA3 only.
- **Names are speakable, not the keycap.** Seven keys on the shipped qwerty layout have an empty or glyph-only cap, so names fall back to the key's own `keyText`.
- **Stale pills needed a real guarantee.** The first version claimed Qt rebuilds the pills every round. It does not when the words are identical: `QQuickRepeater::setModel` returns early on an equal model. The external client caught a held pill still inserting. Every model entry now carries the generation, so each round rebuilds, and a pill's Invoke is refused unless its generation is still live.
- **The window id depended on the application class.** Qt falls back to the class name for an unnamed object, so the shipped app reported `QApplication.alphaOskKeyboard` while the harness reported `QGuiApplication.alphaOskKeyboard`. The application object is now named.
- **Restoring stole the foreground.** Qt restores a minimized window with `SW_SHOWNORMAL` whatever its flags. A client's restore, a restore from another process and the tray's restore code all made the keyboard the foreground window. A native event filter now declines `WM_QUERYOPEN` and restores with `SW_SHOWNOACTIVATE`.
- **A close stranded the keyboard.** `WindowPattern.Close` and the taskbar's Close hid the window while the process kept running, off the taskbar and out of the tree. A close that is not part of a quit now minimizes. Qt 6 cancels a quit if a window refuses to close, so the application object marks a real quit.

## Behaviour changes outside the scanning contract

These affect users who never run a scanner.

- **Taskbar Close now minimizes the keyboard** instead of hiding it to the tray. The title-bar close and the tray's Quit still exit.
- **Restoring the keyboard no longer makes it the foreground window.**
- **Two prediction tests were fixed** in their own commit (`70ac8d1`). They read a lazily built object without building it, and they failed whenever xdist scheduled them apart from the test that built it. Nothing in the prediction engine changed.

## Security

None of this gives a client a capability it lacked. Anything that can reach the window through UIA can already call `ShowWindow` on it, post it a close message, or end the process. What the changes remove is harm from standard, well-meant calls: a restore that moved focus, a close that stranded the keyboard, and targets that stayed live while off screen.

## Verification

- **Live UIA clients.** Every row in the doc's verification table was measured against the real `Main.qml`, with key synthesis recorded rather than sent. The window-state work ran in a harness with its own settings key and config directory, and a control window in each run showed a focus steal could be detected.
- **External client.** `OwenMcGirr/alpha-osk-scan-lab`, a Rust UIA3 client, retested the stale-pill and window-id fixes at `b75ec07`. All 13 of its live checks passed, including an old pill held across an identical round and invoked with no comparison first. That run predates the window-state commit.
- **Tests.** 80 tests across `tests/test_qml_scan_targets.py`, `tests/test_windows_window.py` and `tests/test_keyboard_app.py`. The new guards were mutation-tested: removing each one fails the test written for it.
- **Gate.** `python check.py` passes: ruff, format, mypy for both platforms, and the full suite.

## Test plan

- [x] `python check.py` green
- [x] Guards mutation-tested
- [x] Live UIA clients: enumerate, Invoke a key and a pill, toggle a modifier, stale pill refused, minimize and restore without a focus steal, close minimizes, quit still exits
- [x] External scan-lab retest of the stale-pill and window-id fixes
- [ ] External retest of minimize, restore and state
- [ ] Signed, installed build with UIAccess active
- [ ] Mixed-DPI and multi-monitor alignment
- [ ] A physical switch and real desktop insertion

## Out of scope for v1

As agreed in the issue: the settings, help, snippets, symbols and dashboard windows, secondary-click actions, and held-key commands.

### DIFF
--- a/qml/Main.qml
+++ b/qml/Main.qml
@@ -16,6 +16,13 @@ Window {
     minimumHeight: 200
     color: "transparent"
     title: "Alpha-OSK"
+    // Qt synthesises the window's UI Automation AutomationId from its
+    // objectName, as "QGuiApplication.<objectName>".  Pinning it gives an
+    // external scanner a stable anchor to find this window by, which the
+    // title cannot be (it is user-facing text) and the window class must not
+    // be (Qt generates it, and it changes under us on a Qt upgrade).  See
+    // docs/architecture/UIA_TARGETS.md.
+    objectName: "alphaOskKeyboard"
 
     // Persistent settings — saved automatically on change, restored on launch
     Settings {
@@ -592,6 +599,55 @@ Window {
     // Keyboard layout (data-driven from JSON)
     property var layoutRows: keyboard ? keyboard.getLayoutRows() : []
     property string currentLayout: appSettings.savedLayout
+
+    // ===== External switch-scanning targets =====
+    //
+    // The contract an external scanner (Switchify and any other Windows
+    // assistive technology) reads through UI Automation.  Normative
+    // description, including why each piece is shaped this way, is in
+    // docs/architecture/UIA_TARGETS.md; issue #106 is the conversation.
+    //
+    // The id scheme lives here and nowhere else.  Five surfaces draw
+    // scannable keys and a sixth draws the pills, and a second copy of this
+    // format string is the parallel-blocks failure this project keeps paying
+    // for: a surface that drifted would hand a scanner ids that collide with
+    // another section's, and the scanner cannot tell that from a relayout.
+    readonly property string scanContractVersion: "aosk.v1"
+    function scanTargetId(section, row, idx) {
+        return root.scanContractVersion + "." + section + "." + row + "." + idx
+    }
+
+    // Predictions get a generation rather than a bare slot, so that an id a
+    // scanner has already seen always denotes the same offer.  It is bumped
+    // whenever the pill row is repopulated, including a repopulation that
+    // happens to produce the same words: "the bar was rebuilt" is the event a
+    // stale selection has to be caught by, not "the text changed".
+    property int predictionGeneration: 0
+    onPredictionsChanged: root.predictionGeneration += 1
+    function scanPredictionId(idx) {
+        return root.scanContractVersion + ".pred." + idx + ".g" + root.predictionGeneration
+    }
+
+    // One cheap thing for a scanner to poll.  Everything that can move a
+    // target, change its label, or change whether it is there at all feeds
+    // this string; the scanner reads one property at whatever rate its
+    // overlay needs and takes a full cached snapshot only when it changes.
+    // Geometry is included directly because window drag and resize are the
+    // changes with no other signal behind them.
+    readonly property string scanRevision: [
+        root.predictionGeneration,
+        Math.round(root.x), Math.round(root.y),
+        Math.round(root.width), Math.round(root.height),
+        root.currentLayout, root.compactView ? 1 : 0, root.activeLayer,
+        root.showNumberRow ? 1 : 0, root.showFunctionRow ? 1 : 0,
+        root.showExtraFunctionRow ? 1 : 0, root.showNavigation ? 1 : 0,
+        root.showNumpad ? 1 : 0, root.suggestionsEnabled ? 1 : 0,
+        root.shiftOn ? 1 : 0, root.capsOn ? 1 : 0, root.ctrlOn ? 1 : 0,
+        root.altOn ? 1 : 0, root.winOn ? 1 : 0,
+        root.shiftLocked ? 1 : 0, root.ctrlLocked ? 1 : 0,
+        root.altLocked ? 1 : 0, root.winLocked ? 1 : 0,
+        root.visible ? 1 : 0
+    ].join(".")
 
     // ===== Compact view =====
     // A *view* preference, orthogonal to which letter arrangement is picked:
@@ -1354,6 +1410,24 @@ Window {
     // this back to true to keep the screenshots' corners.
     property bool selfRoundedCorners: Qt.platform.os !== "windows"
     readonly property real windowRadius: selfRoundedCorners ? 10 : 0
+
+    // The one property an external scanner polls.  Its Name is the revision
+    // string above; when that changes, something in the scanning contract
+    // moved and the scanner takes a fresh cached snapshot of the subtree.
+    // That keeps the steady-state cost at one cross-process property read
+    // rather than a walk of sixty-odd elements at overlay frame rate.
+    //
+    // It has to be `visible` to be in the accessibility tree at all (an
+    // invisible item is pruned), so it is a real but empty 1x1 item rather
+    // than `visible: false`.  It draws nothing.
+    Item {
+        objectName: "scanRevisionBeacon"
+        width: 1
+        height: 1
+        Accessible.role: Accessible.StaticText
+        Accessible.name: root.scanRevision
+        Accessible.id: root.scanContractVersion + ".revision"
+    }
 
     // Main background — uses Qt.rgba so only the background becomes transparent
     // while keys and text remain fully opaque
@@ -2528,6 +2602,38 @@ Window {
                                           : root.predPillBorder
                             border.width: predMouse.containsMouse ? 2 : 1
 
+                            // ===== Switch-scanning target =====
+                            //
+                            // The stale-selection guarantee the contract makes
+                            // (an outdated scan selection can never insert a
+                            // word the user was not looking at) rests on object
+                            // lifetime, not on the id below, and that was
+                            // measured rather than assumed: swapping the model
+                            // array makes Repeater destroy and rebuild every
+                            // delegate, so a scanner holding an element from
+                            // the previous round gets ElementNotAvailable and
+                            // its Invoke does nothing.  It holds even when the
+                            // new round produces identical words, because a
+                            // fresh array is a fresh array.
+                            //
+                            // The generation in the id is the *client's* half:
+                            // it lets a scanner notice the change cheaply and
+                            // keeps an id from ever denoting two different
+                            // offers.  Do not "optimise" the model into a
+                            // reused list: that would trade the guarantee for
+                            // a few allocations.
+                            // Named rather than inlined into the attached
+                            // property, for the reason KeyButton's own block
+                            // gives: an attached `Accessible.*` is unreadable
+                            // from PySide, so anything expressed only there is
+                            // untestable.
+                            property string scanTargetId: root.scanPredictionId(index)
+
+                            Accessible.role: Accessible.Button
+                            Accessible.name: modelData
+                            Accessible.id: scanTargetId
+                            Accessible.onPressAction: keyboard.pressPrediction(modelData)
+
                             // Subtle gradient for depth
                             Rectangle {
                                 anchors.fill: parent
@@ -2864,6 +2970,8 @@ Window {
                         objectName: "extraFunctionRowPanel"
                         visible: root.showExtraFunctionRow
                         Layout.alignment: Qt.AlignHCenter
+                        scanSection: "fn2"
+                        scanIdFor: root.scanTargetId
                         keyGroups: [
                             ["F13", "F14", "F15", "F16"],
                             ["F17", "F18", "F19", "F20"],
@@ -2893,6 +3001,8 @@ Window {
                         objectName: "functionRowPanel"
                         visible: root.showFunctionRow
                         Layout.alignment: Qt.AlignHCenter
+                        scanSection: "fn1"
+                        scanIdFor: root.scanTargetId
                         keyW: root.keyW
                         keyH: root.keyH * 0.7
                         keySpacing: root.keySpacing
@@ -2932,6 +3042,8 @@ Window {
                         objectName: "numberRowPanel"
                         visible: root.showNumberRow
                         Layout.alignment: Qt.AlignHCenter
+                        scanSection: "num"
+                        scanIdFor: root.scanTargetId
                         keyW: root.keyW
                         keyH: root.keyH
                         keySpacing: root.keySpacing
@@ -2963,6 +3075,10 @@ Window {
                         Layout.alignment: Qt.AlignHCenter
                         spacing: root.keySpacing
                         property var rowData: modelData
+                        // Visual row order, for the scan-target id below.  The
+                        // inner Repeater's own `index` is the key's slot, so the
+                        // row's has to be captured here before it is shadowed.
+                        property int rowIndex: index
                         property real rowKeyH: rowData.id === "number" ? root.keyH - 4 : root.keyH
 
                         // Equal unit totals are NOT equal pixel widths, and the
@@ -3021,6 +3137,11 @@ Window {
                                 hitMarginH: root.keyHitMarginH
                                 hitMarginV: root.keyHitMarginV
                                 fontSize: kd.fontSize || 16
+                                targetId: root.scanTargetId("grid", rowIndex, index)
+                                // A key with a stateKey is one that can be on or
+                                // off (the modifiers, Caps, NumLock); nothing else
+                                // reports a toggle state to a scanner.
+                                isToggleTarget: !!kd.stateKey
                                 isSpecial: kd.type !== "char"
                                 isActive: {
                                     if (!kd.stateKey) return false
@@ -3231,6 +3352,8 @@ Window {
                 // Flush with the grid on both rails; the panel divides
                 // this between its five rows itself.
                 Layout.preferredHeight: root.sectionHeight
+                scanSection: "nav"
+                scanIdFor: root.scanTargetId
                 roleColors: root.keyRoles
                 // Vertically off `keySpacing`, not `rowSpacing`: this
                 // panel lays its own rows out on it.
@@ -3260,6 +3383,8 @@ Window {
                 objectName: "numpadPanel"
                 visible: root.showNumpad
                 Layout.preferredHeight: root.sectionHeight
+                scanSection: "pad"
+                scanIdFor: root.scanTargetId
                 roleColors: root.keyRoles
                 // Vertically off `keySpacing`, not `rowSpacing`: this
                 // panel lays its own rows out on it.
--- a/qml/components/KeyButton.qml
+++ b/qml/components/KeyButton.qml
@@ -28,6 +28,21 @@ Item {
     property bool isLocked: false
     property bool isWide: false
 
+    // ===== External switch-scanning target (see docs/architecture/UIA_TARGETS.md) =====
+    //
+    // `targetId` is this key's AutomationId, and setting it is what puts the
+    // key in the UI Automation tree an external scanner walks.  A key that
+    // leaves it empty is `Accessible.ignored` and so is absent from that tree
+    // rather than present and unactivatable, which is the property the
+    // contract rests on: presence means invokable.  Callers own the id, since
+    // only they know which section and slot the key occupies.
+    property string targetId: ""
+
+    // Whether this key reports a toggle state to a scanner.  Modifiers and
+    // the lock-style toggles (Caps, NumLock) do; a letter does not, and
+    // claiming otherwise would tell a scanner a letter can be "on".
+    property bool isToggleTarget: false
+
     // How far the hit area reaches past the keycap's own slot, per
     // axis, in pixels.  The gap between two caps is otherwise dead: a
     // click landing in it types nothing and shows nothing, and unlike a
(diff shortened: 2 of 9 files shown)


OwenMcGirr:
The scanning integration looks good in the tests so far. I've published a standalone Windows Rust overlay test app here: https://github.com/OwenMcGirr/alpha-osk-scan-lab

Switchify does not have scanning yet, so this exercises the client side independently. It discovers Alpha through UIA, watches the revision beacon, compares target maps, draws a click-through overlay, and supports linear and row-then-key scanning.

Against commit `0641e0ad8866eaad945518e8d41ed0439a7cf3c2`, all 11 live UIA checks and the native release UI smoke test passed. These covered movement, layout/panel changes, modifiers, key and prediction activation through the recording synthesizer, stale-selection rejection by the client, privacy, and visibility. Foreground was preserved in the recorded activation tests.

Two findings to follow up:

- The QApplication launcher exposes `QApplication.alphaOskKeyboard`. The client accepts that alongside the documented `QGuiApplication.alphaOskKeyboard`.
- The provider-side stale-prediction guarantee did not hold for identical replacement words in this environment. A retained element's generation changed, but invoking that old reference succeeded and produced one synthesis call. Client revision/identity checks rejected an outdated selection, but checking and invoking are separate operations, leaving a race. This needs attention before relying on the lifetime guarantee described in the PR.

Mixed-DPI multi-monitor alignment, physical switch/global-key delivery, real desktop text insertion, and the installed UIAccess build still need testing. The fixture records synthesis calls rather than injecting desktop input.

Reproduction commands and results are in [README.md](https://github.com/OwenMcGirr/alpha-osk-scan-lab/blob/main/README.md) and [VALIDATION.md](https://github.com/OwenMcGirr/alpha-osk-scan-lab/blob/main/VALIDATION.md).

owenpkent:
Yes, I can include it as a layout in the next release. Layouts in Alpha-OSK are plain JSON files in data/layouts, and the keyboard picks them up on its own, so a new one appears in the layout picker next to qwerty, dvorak and colemak with no code change. Its keys come through the same aosk.v1 target ids as every other layout, so nothing on the scanning side changes. Keeping it as a layout also means mouse users keep their default keyboard and switch users pick this one once.

What I need from you is what it should look like. Specifically:

- Key order: alphabetical, frequency-ordered, or qwerty.
- Grid shape: how many rows and columns, and whether every row should be the same length so row-then-key scanning has no ragged edge.
- What is on it: letters only, or digits and punctuation too, and whether those live on a second page reached by a layer key (the way the compact layout does it) or stay on the main grid.
- Where space, backspace, enter and the modifiers sit, and whether they should be grouped so they scan as one row.
- The prediction bar: keep it above the keys as now, and how many pills you want in the scan.
- Anything to leave off entirely, such as the function rows, the navigation cluster and the numpad.

A rough sketch, a screenshot of a layout you already like, or the rows written out as strings would all be enough. If you would rather I propose a first cut and you correct it, say so and I will post one.

OwenMcGirr:
Thanks Owen. I think the current layout is good for us, so there's no need for a separate scanning layout at this stage.

What would help is a way for Switchify or another external accessibility app to minimise/hide the keyboard and recall/show it again without taking focus from the application receiving text.

Could you also expose the keyboard's current visibility state, so a client can check whether it's shown, minimised or hidden when it connects, and track changes afterward? That would let Switchify keep its scanning overlay and keyboard controls in sync, including when the user changes visibility directly in Alpha-OSK.

The recall control and visibility status would need to remain discoverable while the keyboard is minimised or hidden. Ideally this would fit into the documented integration contract alongside the existing UIA targets.

owenpkent:
Thanks for the scan lab. Both findings were real, and both are fixed in b75ec07 on this PR.

**Stale predictions.** You were right, and my claim was wrong. I said a new round always destroys the pills. It does not when the words are identical: Qt's `Repeater.setModel` returns early if the new list compares equal to the old one, so the old pills survived and a held one stayed invokable. My earlier measurement only covered a round where the words changed. The guarantee now rests on two independent things on the keyboard side:

- Every pill in the model carries the round's generation, so no two rounds compare equal and every round rebuilds the pills. An element held from an earlier round can no longer be invoked.
- A pill's Invoke passes its own generation, and the keyboard inserts nothing unless that round is still live. The rebuild should make this unreachable. It is there so a future change that lets a pill survive fails closed.

The id is now built from the pill's own generation, so it no longer changes under a held element as it did in your run. Please don't match on a specific error for the failed Invoke. The earlier UIA3 run reported `UIA_E_ELEMENTNOTAVAILABLE`, while the managed `System.Windows.Automation` client reports "Unsupported Pattern". Keep your pre-Invoke comparison, but the guarantee no longer depends on it.

**Window id.** The id is now `alphaOsk.alphaOskKeyboard`, whichever application class starts the keyboard. Qt builds the id from the application object's name and substitutes the class name when there is none. The application was unnamed, so the shipped keyboard reported `QApplication.alphaOskKeyboard` and my harness reported `QGuiApplication.alphaOskKeyboard`, and I published the harness's. It is now named explicitly. I confirmed both unnamed forms and the new one against a live UIA client. Your client can drop both fallbacks. Nothing has shipped with either old id.

`docs/architecture/UIA_TARGETS.md` is updated to match, including a note on how the stale-pill guarantee was wrong the first time.

When you have a moment, could you rerun your stale-selection check against b75ec07 with an identical round, holding the element and invoking it with no client-side comparison? That is the case your fixture caught and my headless tests could not.

OwenMcGirr:
Retested against b75ec07125addc58896f0835fbb38879d4811d72. Both fixes passed in the scan lab.

For the identical-word round, I retained the old UIA element and attempted activation directly, with no client-side comparison gating it. It produced zero synthesis calls. The different-word replacement case also produced zero calls. In both cases the old identity was no longer readable and Invoke-pattern retrieval failed. We record the returned error but no longer require a particular HRESULT. A current prediction element still inserted successfully through the recording synthesizer.

Discovery now uses only alphaOsk.alphaOskKeyboard, with both old fallbacks removed. The isolated fixture uses your application object name while retaining separate settings. The normal scanner still keeps its pre-Invoke revision and identity checks.

All 13 live UIA checks and the native release UI smoke test passed, including foreground preservation. Formatting, Clippy, the six core unit tests and release build passed too. The fixture recorded synthesis instead of injecting desktop input; signed installed UIAccess, mixed-DPI multi-monitor behavior, physical switch delivery and real desktop insertion remain unverified by these tests.

The scan-lab update is merged: https://github.com/OwenMcGirr/alpha-osk-scan-lab/pull/2

Results and raw reports: https://github.com/OwenMcGirr/alpha-osk-scan-lab/blob/ba6a598b29aabd21120f7110ec4315464746ccdc/VALIDATION.md

owenpkent:
Thanks, noted that the current layout works for Switchify, so there will be no separate scanning layout.

Show, minimize and visibility state are in 69d6e56 on this PR. There is no new control: use the keyboard window's standard UIA WindowPattern. Measuring that against a live client turned up three problems, and those are what the commit fixes.

**The contract**

On the window with AutomationId `alphaOsk.alphaOskKeyboard`:

| To | Call |
|----|------|
| Put the keyboard away | `SetWindowVisualState(Minimized)` |
| Bring it back | `SetWindowVisualState(Normal)` |
| Read the state | `WindowVisualState` |

- **Ignore `CanMinimize`.** It reads `False` because the keyboard draws its own title bar, but the call works.
- **The only states on Windows are `Normal` and `Minimized`.** Every way the keyboard is put away minimizes it, and a minimized keyboard stays in the tree under the same id. If the window is absent, the keyboard is not running.
- **The revision beacon moves on every state change**, including when the user minimizes or restores the keyboard directly. If you already poll the beacon, you will see those changes without polling the window too.
- **A minimized keyboard offers no targets.** Keys and pills leave the tree, and invoking an element you still hold from before the minimize inserts nothing. That refusal is not reported as an error, so check `WindowVisualState` rather than waiting for an Invoke to fail.

**What was broken, now fixed**

- **Restoring took focus.** Qt restores a minimized window in a way that activates it, whatever its flags. A client's restore, a restore from another process, and the tray's restore code all made the keyboard the foreground window, so the next keystroke would miss the application you are typing into. The keyboard now declines that restore and restores itself without activating. Across every run with the fix, a restore never took the foreground. A control window in the same runs did take it, so the check could see a steal.
- **Close stranded the keyboard.** `WindowPattern.Close` and the taskbar's Close both hid the window while the process kept running. It was then off the taskbar and out of the tree, where no client could find it. On Windows, a close now minimizes the keyboard. Its own close button and the tray's Quit still exit. Please don't offer Close as a way to hide it.
- **Minimized keys stayed targetable.** Windows keeps a minimized window's elements in the tree and reports them as onscreen. That is why they are now removed and refuse Invoke while minimized.

**Security**

This gives a client no capability it lacked. Anything that can reach the window through UIA can already call `ShowWindow` on it, send it a close message, or end the process. The changes stop standard, well-meant calls from causing harm: a restore that moved focus, a close that stranded the keyboard, and targets that stayed live off screen.

**Not yet verified**

- The installed, signed UIAccess build. All of this ran on a development build.
- A real click on the taskbar button.
- Mixed-DPI displays.

When you have a moment, could you run minimize and restore from the scan lab with a text field focused in another application, and confirm the field keeps focus? That is the property that matters most here.

The contract in `docs/architecture/UIA_TARGETS.md` now has a section on this, including what was measured.

owenpkent:
Thanks for the retest, and for running the identical round with no comparison gating the Invoke. That was the case that mattered, and zero synthesis calls on both replacement cases is the result I wanted. Keeping your pre-Invoke checks in the normal scanner is right too. The keyboard no longer depends on them, but they cost nothing.

Your run was against b75ec07, and my comment above about minimize, restore and visibility state came after it. That work is in 69d6e56, so none of it is covered by your 13 checks yet. When you next run the scan lab, the checks that matter most are:

1. With a text field focused in another application, `SetWindowVisualState(Minimized)` then `(Normal)`, and confirm the field keeps focus.
2. Hold a key element, minimize, and invoke it. It should insert nothing and leave the tree.
3. Confirm the beacon moves when the user minimizes or restores the keyboard directly, not only when your client does.

On the gaps you listed, the one I can close from my side is the signed, installed UIAccess build. If you want to test against that rather than the branch, say so and I will build a signed installer from this PR for you to install. The mixed-DPI and physical switch checks need your hardware.

qml/components/KeyButton.qml
```diff
@@ -311,6 +387,35 @@ Item {
...
+    // Invoke on Backspace would repeat until the safety timer fired.  One
+    // keystroke, every time.  It goes through `_acceptPress()` for the same
+    // reason a click does, so "behaves like a primary click" stays true of the
+    // debounce as well as the keystroke.
+    //
+    // The flash is not decoration.  A switch user is looking at the keyboard
+    // rather than at the text field, and without it the only feedback that a
+    // scan selection landed is a character appearing somewhere else on screen.
+    function activateFromAssistiveClient() {
+        if (!keyRoot._scanWindowShown)
```
owenpkent: [P1] Refuse retained targets whose panel is no longer visible

This guard checks only the whole window. Numpad, navigation, and function-row delegates remain alive when their ancestor panel is hidden. Reproduction against this head: show the numpad, retain `pad.0.0`'s `QAccessibleActionInterface`, hide the numpad, then call its Press action. `actionNames()` still exposes Press and the recording synthesizer receives `send_text('7')`.

A client that retained a target, or whose visibility check raced with panel closure, can therefore type an unseen key. Removing it from a fresh traversal is insufficient. Check effective target visibility and enabled state at activation time, and use the same state when deciding accessibility presence. Add a test that invokes a retained accessibility interface after hiding its panel, alongside the existing whole-window minimize case.

owenpkent (commented):
## Review of `870f21a`

Four additional findings need attention before the scanning contract is reliable: retained hidden keys remain invokable, the revision beacon misses target changes, NumLock-off targets lack usable names, and assistive activation reuses pointer offsets. Inline comments include focused reproductions and suggested regression coverage.

### Verification

All **80 tests passed** across `tests/test_qml_scan_targets.py`, `tests/test_windows_window.py`, and `tests/test_keyboard_app.py`, running from an exact archive of this head with offscreen Qt and isolated configuration/model directories. The CI run associated with this head also passed.

Additional probes exercised the real QML components and, for the hidden-key case, the retained `QAccessibleActionInterface`. They recorded synthesis calls without sending desktop input. These probes explain why the existing green suite does not establish the missing lifetime and refresh guarantees.

The earlier review discussion about the application-object name and identical-word prediction rounds has already been addressed. I did not find a reason to reopen those resolved technical findings.

### Window lifecycle integration

The new close-to-minimize behavior also makes an existing installer assumption worth resolving: `installer.nsh` treats non-forced `taskkill`/WM_CLOSE as a graceful quit, whereas the PR's own pre-fix measurements say close already left the process alive. That is a pre-existing shutdown integration conflict, not a demonstrated new regression from this PR. An installed-upgrade test should prove that the graceful path runs model saves and modifier cleanup before any forced termination.

### Remaining validation

After the inline issues and current merge conflict are resolved, rerun the focused tests and the full gate. The signed installed UIAccess build, actual Windows UIA transport for these new reproductions, mixed-DPI/multi-monitor alignment, physical-switch delivery, and real target-application insertion remain outside this review's measurements.

qml/Main.qml
```diff
@@ -593,6 +622,73 @@ Window {
...
+        root.visibility !== Window.Minimized && root.visibility !== Window.Hidden
+    readonly property string scanRevision: [
+        root.predictionGeneration,
+        root.visibility,
+        Math.round(root.x), Math.round(root.y),
+        Math.round(root.width), Math.round(root.height),
+        root.currentLayout, root.compactView ? 1 : 0, root.activeLayer,
+        root.showNumberRow ? 1 : 0, root.showFunctionRow ? 1 : 0,
+        root.showExtraFunctionRow ? 1 : 0, root.showNavigation ? 1 : 0,
+        root.showNumpad ? 1 : 0, root.suggestionsEnabled ? 1 : 0,
```
owenpkent, line 725: [P2] Include every target-changing state in the revision beacon

`NumpadPanel.numLockOn` is absent from this string. With the numpad shown, toggling Num leaves `scanRevision` byte-identical while digits become navigation actions and the center key becomes blank/disabled. A scanner following the documented poll-the-beacon protocol keeps its old target map.

A second reproduced case is dictation entering `listening`: the prediction Repeater removes its pills, but neither `predictionGeneration` nor this revision changes. The model also depends on `study.suppressPredictions`, another missing input.

Include NumLock and the complete prediction-presence condition in the beacon, or maintain a dedicated revision bumped whenever the target set, name, or action changes. Add tests that verify both the changed accessible state and the changed beacon, not just traversal after a manual refresh.

qml/components/NumpadPanel.qml
```diff
@@ -115,6 +125,7 @@ Item {
                 borderColor: numpadPanel.borderColor
                 hitMarginH: numpadPanel.hitMarginH
                 hitMarginV: numpadPanel.hitMarginV
+                targetId: numpadPanel._scanId(0, 1)
```
owenpkent, line 141: [P2] Give NumLock-off navigation keys speakable names

These new scan targets rely on `KeyButton._scanName`, which strips non-ASCII glyphs and falls back to `keyText`. The numpad arrow keys have glyph-only caps and never set `keyText`, so with NumLock off the accessible names are empty. The offscreen probe found empty names at pad positions `0.1`, `1.0`, `1.2`, and `2.1`, plus the blank center `1.1`.

This breaks auditory target identification. Bind semantic names for the arrow actions, and omit the disabled blank center from the scan targets, since the documented contract says presence means activatable. Run the name/nonempty-target checks with the numpad visible in both NumLock states; the current default-state coverage misses these controls.

qml/components/KeyButton.qml
```diff
@@ -311,6 +387,35 @@ Item {
...
+            return
+        if (!keyRoot._acceptPress())
+            return
+        keyRoot._visualPressed = true
+        invokeFlashTimer.restart()
+        keyRoot.keyPressed()
```
owenpkent, line 439: [P2] Distinguish assistive activation from a recorded pointer press

`pressDx`/`pressDy` persist from the last mouse press on this key. This invocation emits `keyPressed` without updating their origin, and Main.qml forwards them to `pressKey`/`pressKeyLiteral`. In a focused probe, a key with prior offsets `(0.37, -0.22)` was invoked through this path and the bridge appended `('a', 0.37, -0.22)` to `_word_offsets`.

Those stale coordinates affect live fuzzy matching and are also passed to `observe_press`, so mixed scanner/mouse use teaches pointer bias from a click that never occurred. Carry an explicit non-pointer input source: use the key center for this character's live fuzzy position, keep the offset sequence aligned, and skip pointer-model learning. Merely zeroing the offsets would still train artificial centered clicks. Add a mixed-input regression that proves real pointer samples are retained and scanner activations contribute no learned pointer sample.

owenpkent:
## Addressed in bdc6de7

Rebased onto current `main`; the conflict was the changelog only.

- **P1, a retained target on a hidden panel.** Confirmed and fixed. Hiding a panel leaves every delegate alive, so removing them from a fresh traversal does nothing for a client already holding an interface. `KeyButton._scanActivatable` reads the key's **effective** `visible` (Qt propagates a false parent down, so one test covers window, panel and key) plus `enabled`, and it gates the activation as well as `Accessible.ignored`. `enabled` is what takes the numpad's blank centre key out with NumLock off, per "presence means activatable". Tests: `TestARetainedTargetOnAHiddenPanelCannotType`, invoking the retained `pad.0.0` interface after hiding the panel, paired with the inverse that it still types with the panel shown, plus the disabled centre key. I reverted the guard to the old window-only form and confirmed two of the three fail.
- **P2, the beacon.** Both cases reproduced and both fixed. NumLock is in the string now, and its state moved to `root` so the beacon can see it (the panel raises `numLockToggled` rather than owning it). For the prediction row, the missing input was not the generation but *presence*: `listening` and `study.suppressPredictions` **remove** the pills rather than repopulating them, so the generation never moves. `root.predictionsArePresent` is now the single condition read by both the Repeater's model and the beacon, so they cannot drift apart again. Tests: `TestTheBeaconCarriesEveryTargetChangingState`, each asserting the changed accessible state *and* the changed beacon, with dictation stopping as the inverse.
- **P2, NumLock-off names.** The ten dual-purpose keys bind `keyText` per state, so the four glyph arrows speak as Up / Left / Right / Down. Tests run the numpad in **both** NumLock states and fail on any non-ignored target with an empty name, plus the four arrows by name.
- **P2, the pointer model.** This is the finding I said in #110 I would carry over. `KeyButton.pressFromPointer` travels with the keystroke as a fourth argument to `pressKey` / `pressKeyLiteral` (defaulting True, so every existing caller including the tests is unchanged). A scanner press resolves to the key centre for that keystroke, **keeps its slot** in `_word_offsets` so the sequence still lines up with the word, and calls `observe_press` not at all. `_word_offsets` entries became `(char, offset | None)` for that, which is why the click-position tests changed shape; `positions_for` already read a missing entry as the key centre. As the review said, zeroing the offsets would have trained an artificial dead-centre click, so the source is explicit rather than the numbers overwritten. Tests include the mixed-input pair (a real press still carries its position and still teaches the model) and a QML-side call of the four-argument slot, since PySide overload resolution is the part Python tests cannot see and a failed bind would silently fall back.

`UIA_TARGETS.md` and CLAUDE.md carry all four. On the installer shutdown note: that is a pre-existing `installer.nsh` assumption and an installed-upgrade test, which I have left out of this PR rather than widen its scope.

Full gate green locally (ruff, format, mypy both platforms, pytest).

PR #116, opened 2026-09-13
fix: keep the keyboard where it was left across restarts

owenpkent:
## Summary

The keyboard did not come back where it was left. `Main.qml` restores the saved position in `Component.onCompleted`, then `keyboard_app.py::_apply_window_flags` calls `setFlags` on the already shown window to make it frameless. On Windows, taking the frame off a shown window keeps the old outer rectangle as the new client rectangle, so right after the restore the keyboard jumped 7 px left, 30 px up and grew 15 px wider. `onXChanged` then saved the shifted values, so it crept further every launch (width included).

Measured on the real desktop (150% scaling) with a probe window: restored to (872, 902), sitting at (865, 872) and 15 px wider a moment later. With the flags declared on the QML `Window`, it stays at (872, 902), and `GWL_EXSTYLE` is unchanged (`0x080c0008`, so the taskbar button and topmost bit are unaffected).

## Changes

- `qml/Main.qml`: the root `Window` declares `Qt.Window | FramelessWindowHint | WindowStaysOnTopHint | WindowDoesNotAcceptFocus`, so it is created frameless.
- `src/keyboard_app.py`: `_apply_window_flags` asks for the identical set (adds `Qt.Window`), which Qt treats as a no-op off macOS. macOS still adds `Qt.Tool` afterwards, as before.
- `tests/test_window_position_restore.py`: pins that Main.qml carries the flags before Python touches the window, that the Python call leaves them unchanged, and that the saved position is restored and survives the flag call. The offscreen platform has no native frame, so the shift itself cannot be reproduced headlessly; the cause is what is pinned. Verified the flag tests fail with the QML change reverted.

## Test plan

- [x] `python check.py` passes
- [x] Probe on Windows 11 at 150%: position and size unchanged after `_apply_window_flags`, extended style identical
- [ ] Launch the built app twice without moving it; it should reopen in the same spot


owenpkent (commented):
## Review of `8b0a88b`

No correctness regression found in this diff.

### Implementation

- Declaring the final window flags in `Main.qml` removes the post-restore frame conversion that caused the position and width drift. Including `Qt.Window` in `_apply_window_flags` makes its requested flags match the QML window on Windows and Linux.
- The native Windows styling call still runs after `setFlags`, so the existing no-activation and taskbar behavior continues to be applied. The macOS-only `Qt.Tool` addition is retained.
- Window height remains content-bound. This change adds no persisted height or replacement geometry authority.

### Verification

Ran `tests/test_window_position_restore.py`, `tests/test_windows_window.py`, and `tests/test_keyboard_app.py` from an isolated copy of this commit with PySide6 6.11.1 and the offscreen platform: **34 passed**. The CI run associated with this head also passed.

The new tests establish that the window starts with the required flags, the Python flag request is identical, and saved coordinates survive that call. Their stated limitation is accurate: an offscreen window has no native frame, so these tests do not independently reproduce the Windows frame shift. The two-launch packaged-app check remains useful before release, including persisted width as well as position.

GitHub currently reports a merge conflict for this PR. Resolve it while preserving the equality of the QML and Python flag sets, then rerun the focused tests on the resolved commit.

PR #119, opened 2026-09-13
feat: stop right-click on F-keys opening the key editor

owenpkent:
Right-clicking an F13-F24 or F1-F12 key no longer opens the key action editor; it does nothing. Settings -> Function Keys is now the only way to program a key.

- Removed onKeyRightPressed and the editFn property from FunctionRow.qml, and its two bindings in Main.qml
- Help text and CLAUDE.md updated
- New test right-clicks every key on both rows and checks the editor stays closed and nothing is typed (confirmed failing against the old QML)

Test plan: python -m pytest tests/test_qml_function_row.py (all pass)


owenpkent (commented):
## Review of `2855600`

No correctness regression found in this diff.

### Behavior traced

- Both function-row instances lose the editor callback, and `FunctionRow` no longer handles `keyRightPressed`. `KeyButton` sends right-button presses down that separate signal path, so removing this handler does not turn a right-click into a function-key activation.
- Primary activation still uses `_activate`, preserving ordinary F-key presses and assigned shortcut/text actions.
- The Settings -> Function Keys editing route and its return-to-settings behavior remain intact. Help text and the project documentation describe the resulting interaction consistently.

### Verification

Ran `tests/test_qml_function_row.py` and `tests/test_key_actions.py` from an isolated copy of this commit with PySide6 6.11.1 and the offscreen platform: **54 passed**. The CI run associated with this head also passed.

The new regression test exercises the right-click signal on both rows, checks that the editor stays closed, and verifies that no synthesized key was sent. The surrounding tests cover the retained left-click activation and settings editing flow. This is automated QML/signal verification; physical mouse input on the desktop was not exercised.

PR #120, opened 2026-09-15
fix: keep typed text out of the Linux synthesizer's error logs

owenpkent:
## Summary

- On Linux, `_run` logged the entire xdotool/ydotool command at ERROR on a timeout or any other failure. For `xdotool type` the arguments are the text being typed, so a stalled tool wrote a transcript of typing, password fields included, into `alpha-osk.log`, the file users attach to bug reports. Privacy mode cannot gate this: it deliberately still types.
- `send_key` logged the key name at WARNING when no tool was installed, and the bridge's chord path hands it the typed letter. On macOS, the no-keycode warning logged a chorded key by its character the same way.
- Failures now log the tool, the subcommand, and the error class and errno only, never the arguments or `str(exc)` (`TimeoutExpired` quotes the whole argv). The pre-fix log purge gains a second generation behind its own sentinel, applied on Linux and macOS, so the records already on disk are removed once on upgrade. Windows is not charged: its synthesizer had no typed-content site, so its logs never held these records.

## Related issue

Item 1 of the architecture review in #109, which confirmed the defect and said it should ship on its own.

## Type of change

- [x] Bug fix

## Test plan

- [x] `python check.py` passed through the pre-push hook: ruff, formatting, mypy for Linux and Windows, and pytest.
- [x] Added `tests/test_platform.py::TestLinuxDiagnosticsNeverCarryTypedText` (a timed-out `type`, a failure whose exception message quotes the command, `replace_text` on ydotool, `send_key` with no tool, an empty command) and `TestMacOSDiagnosticsNeverCarryTypedText`. Each negative case is paired with its positive half, that a record still exists and names the tool and subcommand, so a fix that deleted the logging outright fails too.
- [x] Added `tests/test_keyboard_app.py::TestTheSecondLogPurge`: a Linux or macOS install that ran the first purge purges again exactly once; a macOS log holding the old `No keycode for 'é' ...` record is removed; a Windows install that ran the first purge keeps its logs; an install that never purged still purges on every platform; the default platform is the running one.
- [ ] Not exercised on a live Linux desktop against a real stalled xdotool, nor with real Quartz events.

## Accessibility check

n/a: no change to keystroke timing, repeat behaviour or visual feedback. The failure branches log differently and still degrade to a dropped keystroke, as before.

## Notes for reviewers

- Privacy-related change, called out as CLAUDE.md asks: it changes what `alpha-osk.log` can contain and deletes existing logs once on Linux and macOS.
- The second purge covers exactly the platforms whose synthesizer carried a redacted site (`_LOG_PURGE_2_PLATFORMS`). The first push scoped it to Linux alone and the review caught the macOS omission. Generation 1 still applies everywhere, for an install that skipped every intermediate build.
- `_error_name` keeps the class and errno, so a missing binary still reads `FileNotFoundError (errno 2)`. The DEBUG `_log_send` record is unchanged and remains the sanctioned place for the full command.
- Windows was checked as well: its only key-name warning fires for multi-character names, which a typed character never is.
- Docs updated in the same branch: the diagnostic-log bullets in CLAUDE.md, `GOTCHAS.md`, the `_run()` entry in `PLATFORM_ARCHITECTURE.md`, and a CHANGELOG entry under Unreleased.


src/keyboard_app.py
```diff
@@ -378,29 +378,48 @@ def _migrate_legacy_compat_settings() -> None:
...
+    tests can exercise every branch on one machine.  Returns the number
+    of files removed.  Never raises: a failure here must not stop the
     keyboard from starting.
     """
-    sentinel = config_dir / _LOG_PURGE_SENTINEL
-    if sentinel.exists():
+    if platform is None:
+        platform = CURRENT_PLATFORM
+    generations = [config_dir / _LOG_PURGE_SENTINEL]
+    if platform == "linux":
```
owenpkent: [P2] Purge the macOS records redacted by this same change

Generation 2 is restricted to Linux, but `src/platform/macos.py` also had a typed-content warning that this PR fixes. The real bridge's Ctrl/Alt/Win character path reaches `send_key(key.lower(), modifiers=...)`; an unsupported character such as `é` has no macOS keycode. A mocked probe on the base commit emitted `No keycode for 'é' ...` at WARNING. The PR head correctly omits the character.

A macOS installation that already ran generation 1 can therefore contain these pre-fix records. With its first-generation sentinel and an existing log, `_purge_pre_fix_logs(..., platform="macos")` returns 0 and leaves the log untouched. The new test explicitly preserves that outcome, so the assertion that macOS logs never held typed records is incorrect.

Apply a new purge generation to the affected macOS profiles as well as Linux. Add a macOS regression with the old marker and a synthetic no-keycode warning, checking that the log and rotations are removed once and new logs survive the next launch. Keep the Windows preservation case and update the platform-scope statements in the PR description, changelog, and docs.

owenpkent (commented):
## Review of `bb36c4b`

One privacy-cleanup gap remains: this PR redacts a macOS warning that could contain typed characters, but the new purge generation applies only to Linux. The inline comment includes a before/after logging reproduction and the retained-log result.

### Redaction and lifecycle checks

- Traced every internal Linux `_run` call. The first two arguments are the fixed tool and subcommand, so `_describe` does not include typed arguments on those paths. `_error_name` avoids exception messages and preserves only the class and integer errno. Timeout, general failure, missing-tool, and replacement paths continue to produce useful diagnostics.
- Verified that macOS's changed warning removes the key itself. With Ctrl active, the bridge can pass an unsupported Unicode character to `send_key`; the base commit logged that character and this head logs only its length and modifiers.
- The purge still runs before the rotating log handler opens. On Linux, an existing first-generation marker triggers the second purge once, and subsequent launches preserve newly written diagnostics.

### Validation

Ran the exact PR head from isolated source archives: **92 platform tests and 23 app-level tests passed, 115 total**. Additional mocked probes compared the macOS warning on the base and head and checked purge behavior with an existing first-generation marker. `git diff --check` passed.

At review time, [CI run 35032343893](https://github.com/owenpkent/alpha-osk/actions/runs/35032343893) was still in progress; lint, type checks, and OSV had passed. The full CI result remains pending.

Verification ran on Windows with synthesized input recorded or mocked. Real Quartz events, a live Linux desktop, and actual stalled xdotool/ydotool processes were not exercised.

src/keyboard_app.py
owenpkent: Agreed, and fixed in d699a3c. Generation 2 now applies to Linux and macOS through `_LOG_PURGE_2_PLATFORMS`, which is exactly the set of synthesizers that carried a redacted site; Windows stays excluded because its synthesizer had none. Added the macOS regression with the first-generation marker and a log holding the old `No keycode for 'é' and modifiers ['ctrl']` record: removed once, and a post-fix log survives the next launch. The Linux-only claim is corrected in the constants' comments, CLAUDE.md, GOTCHAS.md, the changelog and the PR description.

PR #128, opened 2026-09-16
feat: expand vocabulary to 83k words with compact prediction indexes

owenpkent:
## Summary

- Expand the base vocabulary from 18,989 to 83,307 words: 64,291 filtered ESDB size-60 additions with a minimal base prior, plus 27 curated care, accessibility, and software terms. Include a checksum-pinned generator, source manifest, exclusions, and the full license.
- Pack the SymSpell deletion and fuzzy prefix indexes into contiguous buffers with mutable learning overlays. Preserve ordered lookup behavior and build the prefix index during initialization.
- Avoid scoring the entire base vocabulary on each keystroke by retaining every personal, corpus, and context candidate plus the top matching base candidates. Preserve skipped-apostrophe matching and use deterministic lexical score ties.

## Related issue

No linked issue.

## Type of change

- [x] New feature
- [x] Other: prediction-engine storage and candidate-search optimization, research documentation, and benchmark tooling

## Measurements

Fresh-process fuzzy-index measurements before integration with current main:

| Configuration | Base words | Retained allocations |
|---|---:|---:|
| Original indexes | 18,989 | 44.12 MiB |
| Packed indexes | 18,989 | 12.00 MiB |
| Packed indexes, expanded vocabulary | 83,307 | 48.79 MiB |

These figures exclude the rest of the app. Expanded index construction took about 2.48 seconds with a 122.87 MiB temporary allocation peak. Initialization now pays the prefix-build cost before typing.

The research document records primary-source Gboard findings, algorithm alternatives, reproducible commands, ordered parity checks, isolated scorer timings, and AAC quality results with their limits:
[Research and measurements](https://github.com/owenpkent/alpha-osk/blob/feat/vocab-next-release/docs/research/VOCABULARY_MEMORY.md).

After integrating main at `f0cdb5a`, fresh-model AAC quality runs reproduced 49.6%/45.6% KSR on dev and 50.7%/46.9% on test (clean/slip). The aggregate gains over the original 18,989-word baseline remain below the project's 1.4-point noise yardstick.

## Test plan

- [x] Full local `check.py` gate on `e7fde34`: Ruff, formatting, mypy for Linux and Windows, and pytest.
- [x] Added tests for packed-index parity/collisions/Unicode, mutable overlays, expanded-vocabulary lifecycle, deterministic ties, and candidate pruning against an independent full scorer.
- [x] Separate lint and format checks for the generator and benchmark scripts.
- [x] Windows backend verification with temporary models.
- [ ] Interactive `python run.py` or packaged-release smoke test.
- [ ] Linux and macOS runtime verification.

## Accessibility check

- [x] No new requirement for fast or precise input.
- Preserve main's fixed five-candidate short-prefix rescue floor, skipped-apostrophe matching, and faster personal learning.
- Index preparation happens at startup/rebuild rather than on the first typed prefix.
- Repeat timing, input precision, and visual feedback are unchanged.

## Notes for reviewers

This changes the prediction engine. Please review the pruning proof and fallback conditions, corpus-prior and contraction coverage, deterministic score ties, and immutable-index/mutable-overlay boundaries.

The vocabulary list supplies coverage, not conversational frequency evidence. Reviewed exclusions are not an exhaustive content classification. All generated additions carry one base count; personal history and model schemas are unchanged.

The existing packaging includes the data directory. No build/signing or telemetry behavior changes are included. Startup time and temporary memory are the main measured tradeoffs.
Final integration: merged main at 3a9347f, preserving its UI Automation support and all vocabulary changes. The only conflict was in tests/test_fuzzy_prefix_beam.py; the resolution retains main's built_beam helper and the vocabulary regression tests.

After resolution, Ruff, formatting, mypy for Linux and Windows, all tests changed by the UI Automation update, and conformance checks passed. The full check.py gate and stable AAC results above apply to e7fde34. Completed checks were not repeated during finalization. Details are recorded in docs/research/VOCABULARY_MEMORY.md.


owenpkent:
## Review

Substantial piece of work, and the packed index structures are the interesting part of it. Four things block merging, and I verified each of them myself rather than taking them on trust. A further set of leads is listed at the end, explicitly marked as unverified.

---

### 1. The shipped vocabulary now contains ethnic slurs, and they are unavoidable predictions

This is the one that has to be fixed before anything else is worth discussing.

`data/explicit_exclusions.txt` is an **opt-out** filter of 69 stems, applied to a 64,291-word import. Anything the list did not anticipate ships. Several slurs and explicit sexual terms did not get anticipated, and because they are rare strings they are often the *only* completion of their prefix:

```
rg -N -i "^kik" data/english-expanded.txt     # two matches, both the slur
rg -N -i "^dyk" data/english-expanded.txt     # two matches, both the slur
rg -N -i "^anu" data/english-expanded.txt     # two matches
rg -N -i "^retar" data/english-expanded.txt
```

I have deliberately not pasted the words into this comment; run the above. The point is the prefix behaviour: three keystrokes produce a slur as the **sole** prediction pill, because nothing else in 83,307 words shares that prefix. `main`'s base is `google-10000-english-usa-no-swears.txt` plus a supplement CLAUDE.md describes as "filtered for explicit content", so this is a regression against a property the project had.

On a keyboard that one person depends on daily and screen-shares, this is the worst output the prediction bar can produce. An opt-out stem list cannot get this right; the filter needs to be an allow-list, or the import needs a curated source that has already done the work.

### 2. CI is entirely red, and it is not flaky

All **eight** shards, both operating systems, hit the 20-minute job timeout on run `35128650878`, having reached about 56% of the suite:

```
Test (ubuntu-latest  0..3/4)   20m16s  20m17s  20m51s  21m12s   all fail
Test (windows-latest 0..3/4)   20m46s  20m09s  21m00s  20m08s   all fail
```

Cause, measured with the same script in both trees on this machine:

```
KeyboardBridge()   main 0.79 / 0.79 / 0.82 s     this branch 3.65 / 3.71 / 3.75 s
```

A 4.6x regression, paid by every test that constructs a bridge and by every launch of the application. In the CI logs individual QML fixtures go from roughly 1 s to roughly 24 s each.

Note this runs directly against the reasoning already in the file: the comment above `hybrid_predictor.py:192` removed a duplicate `set_frequencies` call precisely because half a second "paid on every launch, and on every one of the ~1300 tests that builds a bridge" was too expensive. This adds several seconds back.

Raising `timeout-minutes` is not the fix. The cold-start cost is a user-facing regression in its own right: the window now blocks for ~3 s before it is usable.

### 3. The persisted model file triples, and that makes finding 1 permanent

Measured after `load_base_dictionary()` + `save()`:

```
main          18,989 unigrams    ngram_model.json    471,874 bytes
this branch   83,307 unigrams    ngram_model.json  1,488,354 bytes
```

That file is written on every auto-save-on-exit and travels in the Data Backup archive.

The consequence worth spelling out: `save()` persists the whole merged unigram table and `load()` restores it wholesale, so **a word withdrawn from the shipped list in a later release cannot be withdrawn from existing users**. Fixing finding 1 by adding stems to `explicit_exclusions.txt` and regenerating would drop them from `_base_unigrams` but leave them in every saved `unigrams`, where they keep passing the validity gate and keep entering the fuzzy dictionary, until the user runs Clear Learned Data. Whatever the content fix is, it needs a migration path on load.

### 4. A corrupt data file stops the keyboard from starting

`src/prediction/ngram_predictor.py:610` catches only `OSError`. `path.open(encoding="utf-8")` iterated line by line raises `UnicodeDecodeError`, which is a `ValueError`, on a truncated or corrupted `data/english-expanded.txt`: a partial install, a bad download, a disk error. That escapes `_load_extra_vocabulary`, `NgramPredictor.__init__`, `HybridPredictor.__init__` and `KeyboardBridge`, and the keyboard never opens.

`_load_proper_nouns`, immediately below it in the same file, catches bare `Exception` for exactly this reason, as does every other loader here. The new loader also has no entry or byte cap, unlike every other store in the project.

---

### Unverified leads

These came out of the same review pass but I have **not** confirmed them; treat them as things to check, not as findings.

- `hybrid_predictor.py:911` - `enable_vocabulary_pack` looks like the one hybrid path that did not get the eager `prepare_prefix_index()` its two siblings got, which would leave the packed prefix index to rebuild lazily on the next keystroke (claimed 0.56 s).
- `hybrid_predictor.py:971` - `_rebuild_fuzzy_dictionary` is reached from `unprefer`, i.e. a single dashboard tag click, and would now block the UI for seconds rather than the "about half a second" the docstring and CLAUDE.md both still claim. Those two figures are stale either way and should be updated with this PR.
- `packed_deletes.py:55` - `build` appears to call `variants_for(word)` once in the counting pass and again in the fill pass, generating the whole deletion neighbourhood twice.
- `ngram_predictor.py:975` - `_can_prune_base_candidates` walks every corpus-prior word to test a condition that may be unreachable, on the keystroke path.
- `ngram_predictor.py:1050` - `_prefix_successor` looks computed twice for the same prefix, and `cache_result` is gated on `not partial`, leaving one-letter prefixes uncached.
- `prefix_beam.py:109` - `update_word` records overlay children unconditionally, which would turn `children()` into a set union plus sort on the hottest prefixes.
- `packed_deletes.py:118` - encodes the lookup key before probing, where the sibling `PackedPrefixes.find` defers the encode until a hash match.
- `scripts/bench/prefix_memory.py:420` - parses worker stdout as JSON before checking the return code, and exempts the equivalence worker from the return-code check entirely. Same shape in `vocab_memory.py` and `vocab_growth.py`.
- The four new bench scripts duplicate the same worker harness, git-extraction and temp-dir plumbing; the defect above exists in three of them at once.

### How this was checked

Measurements are on a detached worktree at `116d43f` against `main` at `6a1a6de`, same machine, same script in both trees. CI figures are from run `35128650878`. Nothing here was measured on the live model directory.

owenpkent:
### Correction and a number for finding 2

My local suite run finished after I posted, and it sharpens finding 2 in a way that changes where to look for the fix.

**Nothing on this branch fails.** The whole suite passes. It is simply slower, by a consistent factor. Identical command (`pytest -n auto -q --no-header`), same machine, back to back:

```
main         6a1a6de    116.7 s    all pass
this branch  116d43f    389.4 s    all pass
                        3.34x
```

So "CI is red" above should not be read as "tests are failing on this branch". They are not. Each CI shard reaches roughly 56% of its slice and is killed by `timeout-minutes: 20`, because a 4-vCPU runner with four xdist workers applies that same 3.34x to a job that already takes 6 to 9 minutes.

That reframes the remedy. There is no broken test to find; the cost is the 83k dictionary being built per fixture, which the `KeyboardBridge()` figure locates precisely (0.80 s to 3.70 s per construction). The productive directions are making the dictionary load lazily, sharing one prepared index across tests that do not mutate it, or both. Raising the CI timeout would hide the same cost that users pay at every launch, so it is the one thing that should not be the fix.

Worth noting the local factor (3.34x) is milder than CI's (past 3.0x on top of a job that was already near half the limit). If you get the construction cost back under roughly 1.5 s, both the shard timeouts and the ~3 s cold start go away together.

owenpkent (commented):
## Verdict: do not merge

Changes-requested is not available on your own PR, so this is the verdict as a review comment.

**Blocking, in priority order:**

1. **Vocabulary content.** Slurs are reachable as the *sole* prediction from a three-letter prefix. This alone disqualifies the branch regardless of everything below, and the opt-out stem list is the wrong shape to fix it. An allow-list, or a source that has already done the curation, is what this needs.
2. **Content fix must carry a load migration.** `save()` persists the merged unigram table and `load()` restores it wholesale, so regenerating the shipped data cleans new installs only. Without a migration, existing users keep the words.
3. **Cost.** 3.34x suite time and 4.6x `KeyboardBridge()` construction, which is what times out all eight shards and adds ~3 s to cold start. Not a broken test; a per-fixture dictionary build.
4. **`_load_extra_vocabulary` catches only `OSError`**, so a corrupt data file stops the keyboard starting.

Items 1 and 2 are the ones that make this a rework rather than a tidy-up: they change what the feature ships and how it upgrades. 3 and 4 are tractable in place.

The packed index work underneath is worth keeping. My suggestion would be to split it: land `packed_prefixes` / `packed_deletes` and the bench harness against the existing 19k vocabulary, where they can be reviewed on their merits and measured without the content question, and let the vocabulary expansion be its own PR with a curated source.

owenpkent:
Reviewed against `main` at 3a9347f. I traced the engine changes and measured the costs locally rather than inferring them from CI.

## The engine work is sound

The candidate pruning is well proven. For a base-only word the score is a strictly monotonic linear function of raw base frequency, so "top-n by base frequency" is identical to "top-n by score", and every non-base-only candidate (trigram, bigram, user or corpus contribution) is unioned in unconditionally. Pruning can therefore only drop pure-base words that could not have made the cut anyway. The apostrophe-optional match is correctly mirrored through a second sorted index under the same guard as `_matches_partial`. The packed-buffer offset arithmetic, the UTF-8 handling and the immutable-buffer/mutable-overlay split all check out, and the `(-score, word)` tie-break is a genuine determinism fix over the previous set-iteration order.

I did not find a case where pruning changes which word holds a pill slot.

## CI cannot currently pass

All eight shards hit the 20 minute job timeout. They were at 56% and still passing tests steadily, so this is a slowdown rather than the hang fixed in #130. Locally, `tests/test_hybrid_predictor.py` goes from 1m06s to 5m18s.

The proximate cause is that `prepare_prefix_index()` is now eager at construction, so every predictor instance pays the full index build rather than the first typed prefix paying it once per process.

## The cost belongs to the vocabulary, not the packing

| configuration | construct | `tomo` pills |
|---|---|---|
| `main` | 0.53s | `tomorrow, tonight, timothy, finish, room` |
| this PR, packed indexes, original vocabulary | 0.85s | identical to `main` |
| this PR, packed indexes, expanded vocabulary | 3.45s | `tomorrow, tomographic, tomography, tonight, tomorrows` |

The packed indexes on their own cost about 0.2s over `main`, and keep the 44.12 to 12.00 MiB memory win with byte-identical predictions. The vocabulary expansion accounts for about 2.6s of the 2.8s startup regression and for all of the ranking churn.

## Two quality findings on the vocabulary half

1. Across a 6,084 prefix sweep, 7.1% change their slot-1 pill and 17.0% change their top three. Pill position is muscle memory, and the promoted words are spell-checker coverage (`aigrette`, `ahchoo`, `bhaji`, `tomographic`) rather than conversational vocabulary.

2. The dead-prefix rescue shrinks: 298 to 266 dead two-letter runs, and 15,153 to 14,579 dead three-letter runs. Expanding the wordlist makes previously dead prefixes live, which switches the rescue off for them. `aig` currently rescues to `with`; with the expansion it returns `aigrette`. That rescue came from a user report, so this is the trade worth weighing most carefully.

The 298 figure matches the number already recorded in CLAUDE.md, which is what gives me confidence the harness is measuring the right thing.

Set against those costs, the PR's own AAC numbers put the aggregate gain below its stated 1.4 point yardstick.

## Suggested path: split this PR

`extra_vocabulary` is a single optional field on the language profile, so setting it to `None` lands the packed indexes now (real memory win, no ranking change, CI green) and moves the vocabulary question to its own PR where it can be judged on its own evidence.

## Smaller note

`_fuzzy_frequencies` now inlines `chain(unigrams, _base_unigrams, _corpus_unigrams)` instead of going through `vocabulary()`, and drops the comment explaining why that choke point exists. It is set-equivalent, so not a bug, but it reaches into two private attributes and leaves the three-layer union stated in two places. An ordered enumeration on `NgramPredictor` would keep both the choke point and the ordering fix.

One unverified item worth a look: `_top_unigrams_with_scores` appears to still scan the whole merged unigram table, which the expansion inflates. It looks reachable only when `bigrams[SENTENCE_START]` is empty, so it may be a narrow edge case, but I did not confirm it.

owenpkent:
Opened #134 with the packed-index half of this branch, on the split both reviews recommended.

It carries `packed_deletes`, `packed_prefixes`, the SymSpell and `PrefixIndex` changes, the candidate pruning and the two bench scripts, against the existing 19k vocabulary. Left here: the wordlist, its generator, manifest, licence and exclusions, and the `extra_vocabulary` profile field, which I dropped rather than left unused on this repo's own rule that a field nothing reads is worse than one that does not exist yet. It comes back with the curated list.

Measured on #134, best of three, vocabulary held constant so it is not a variable:

| | main | #134 |
|---|---:|---:|
| retained memory | 73.5 MiB | 41.5 MiB |
| to first pill | 0.54 s | 0.75 s |
| first prediction | 0.093 s | 0.007 s |
| full suite | ~124 s | 142 s |

The 6,084-prefix sweep is byte-identical to main there, against 1,500 differing prefixes here, and the suite is 15% slower rather than 3.34x, so it fits the shard timeout this branch exceeds.

That leaves this PR as just the vocabulary question, which is where the blocking findings are: the content problem, and the load migration it needs because `save()` persists the merged unigram table and `load()` restores it wholesale. I confirmed the content finding independently before acting on it: of 20 probe terms, 11 are in `data/english-expanded.txt` and 7 are reachable in the top five from a three-to-five character prefix. I could not reproduce the "sole prediction" phrasing (none of mine were the only pill), but the substance holds.

Worth noting the fourth item does not travel to #134 either, since the loader went with the wordlist: `_load_extra_vocabulary` catching only `OSError` means a corrupt data file stops the keyboard starting, and that wants fixing wherever the loader lands next.

owenpkent:
Revived on top of current main, with the three remaining blockers addressed. Content is settled per your decision, so the wordlist ships unfiltered and the new setting governs it instead.

**Cost.** Profiling put 67% of construction in SymSpell's deletion index (about 4.3 million deletion variants) and 19% in the packed prefix index. Both are built from a word-frequency mapping and neither mutates afterwards, which `#134`'s packed representation is what makes true, so they are now shared across instances through a process-level cache keyed on a BLAKE2b digest of the exact input. Python's own `hash` would not do: a collision would not raise, it would quietly answer one vocabulary's predictions from another's index.

| | before | after |
|---|---:|---:|
| construction 1 | 2.78 s | 2.78 s (cold) |
| construction 2+ | 2.78 s | 0.43 s |
| full suite | 389 s | **114 s** |

Main's own suite is ~124 s, so this branch is now *faster* than main despite carrying four times the vocabulary, and a warm construction beats main's 0.74 s.

**`_load_extra_vocabulary`.** Now catches every exception, not just `OSError`. It runs inside `__init__`, so anything escaping it stops the keyboard starting, and the user whose only input device is this keyboard cannot then repair the file. Paired tests: the corrupt-file case fails with exactly `UnicodeDecodeError` against the old catch, and a missing-file case beside it stops a loader that swallowed everything by doing nothing from passing both.

**The filter.** *Smart Typing -> Suggestions -> Filter Explicit Words*, default on. It filters what the bar volunteers, not the vocabulary: flagged words stay in the dictionary, stay typable and stay learnable, and personal vocabulary outranks the filter, so a word you have typed comes back.

Generating the list surfaced something worth knowing. The build-time exclusion list was written to cover swearing, so **the shipped wordlist contains slurs but no common profanity**. The 37 slur stems are that gap. Restoring profanity to the vocabulary means regenerating without the exclusion, which is a separate decision I have not taken.

Matching is stem-plus-closed-suffix, never substring, which is the Scunthorpe rule; `spook` is deliberately not a stem because it takes `spooky` and `spooked` with it. Verified over the 6,084-prefix sweep: no flagged word surfaces anywhere with the filter on, four do with it off, and zero of 37 ordinary words are flagged. The list is cross-checked against the shipped no-swears frequency list as ground truth.

Full `check.py` gate green. One nice incidental: the inventory test from #133 caught the two new public methods on `HybridPredictor` and failed until they were classified, which is what it was built to do.

PR #132, opened 2026-09-16
docs: qualify AAC benchmark and pointer-learning claims

owenpkent:
## Summary

The whitepaper overstated synthetic pointer results as evidence of per-user learning and treated the difference between two corpus splits as a statistical noise threshold.

- Qualify the historical results and keep prediction-bar KSR, correction accuracy, WER/CER, latency and user effort separate.
- Add a sourced audit of Will Wade's AAC dataset, Correct-A-Sentence, Text Slinger and VelociTap, with adapter requirements, test cases and a staged benchmark plan.
- Defer the study protocol's unsupported pointer-precision analysis and document the remaining pointer, blacklist-rehabilitation and analytics isolation requirements before enrolment.

## Related issue

No linked issue. Follow-up to the AAC dataset and correction-work audit.

## Type of change

- [x] Docs only

## Test plan

- [x] Reviewed the claims against source material and the current benchmark, pointer model and study paths.
- [x] `git diff --check` passed.
- [x] Verified the applied files match the reviewed patch and checked local Markdown links.
- [x] Checked whitespace and the repository's punctuation convention.

The application test suite and benchmarks were not rerun for these documentation changes. Historical measurements retain their original configuration and limitations.

## Accessibility check

N/a: documentation only; no interaction changes.

## Notes for reviewers

The study's incomplete state isolation is documented here and still requires an implementation fix before enrolment. This PR changes no prediction, build or telemetry code, consent/export schemas, or live learned data.


owenpkent:
## Review

This is good and mostly careful work. The rewrites do the thing that is hard to do to your own paper: they separate what was simulated from what was measured on a user, stop reading ablation deltas as causal, and rename §8.5 from "Personalisation" to "In-domain text exposure", which is what that experiment actually is. The `--learn-half` oracle row in particular was being asked to carry an argument it could not carry, and now it is not.

One thing blocks it, and it is a same-sitting fix rather than a reason to reject.

---

### The noise floor is removed from the whitepaper and left standing in four other places

§8.2 previously set the convention, "we treat differences under about 1.4 points as not meaningful", and this PR replaces it with the correct statement that a single split-to-split gap is descriptive of corpus sensitivity and is not a confidence interval.

Agreed. But the convention it retires is asserted, and reasoned from, in four files this PR does not touch:

- **`CLAUDE.md:121`** - "treat anything under 1.4 points as noise, which is what the two AAC splits disagree by with nothing else changed". This is the authoritative onboarding doc, so after this merges it instructs the next contributor to apply a rule the whitepaper now says is unsound.
- **`docs/architecture/PREDICTION_NOTES.md:78-82`** - "**The gap between `aac-dev` and `aac-test` is 1.4 points** ... That is a usable noise floor: a one-point difference is not evidence even on these sets". This one directly contradicts the new §8.2 text.
- **`docs/architecture/FUZZY_RECOGNITION.md:258`** - reasons from "small aggregate gains are below the benchmark's noise floor" to justify a design decision.
- **`docs/research/PHILOSOPHY.md:123-125`** - the worst of the four, because it *quotes* the whitepaper: "The four user-selectable merge strategies are separated by less than the noise floor. The setting stays, and **the whitepaper says plainly that we cannot demonstrate any choice beats the default.**" This PR deletes the sentence being cited, so the citation dangles.

Either the convention survives as an explicitly-labelled working heuristic ("we use 1.4 points as a rule of thumb for deciding what to chase, and it is not an inferential claim"), or it goes everywhere. What it cannot do is go in the paper and stay in the instructions.

`PHILOSOPHY.md` needs a wording change regardless, since it attributes a sentence to the whitepaper that will no longer be there.

### Smaller points

- **A rationale is lost in §8.3.** The old merge-strategy paragraph explained *why* the setting stays despite the flat aggregate: "it is defensible on the grounds that different strategies fail differently on individual queries". The new text drops that and keeps only the statistical disclaimer. The disclaimer is the improvement; the reason the setting exists is still worth a clause, and `PHILOSOPHY.md` currently carries it alone.
- **§8.4's cross-reference.** §8.2's "Model state" paragraph now says "Section 8.4 enables online pointer adaptation for one condition, and section 8.5 deliberately exposes the language model to additional text", where the old text said "Section 8.5 is the one exception". Worth double-checking that 8.4 really was an exception to the cold-start rule in the original runs, rather than that sentence being widened on inference.
- **The audit doc pins an unmerged PR.** `AAC_BENCHMARK_EVIDENCE.md` references "Vocabulary PR #128 at `116d43f`". That is accurate today and will read oddly once #128 lands or is rebased. A sentence rather than a SHA would age better.
- **Revision line.** "2.1 (16 September 2026) ... evaluation claims clarified without rerunning benchmarks" is exactly the right thing to put there.

### What I checked

Cross-references were grepped against `main` at `6a1a6de` for the files this PR does not modify, which is where all four of the above live. I did not re-run any benchmark, which is consistent with the PR's own stated scope.

owenpkent (commented):
## Verdict: merge, after one same-sitting fix

Approve is not available on your own PR, so this is the verdict as a review comment.

The substance is right and this should land. It does the hard thing to your own paper: separates simulated from measured, stops reading ablation deltas as causal, and renames 8.5 to what that experiment actually is.

**One thing to do before merging**, because it is the failure this project explicitly warns about elsewhere (updating cross-references together): the 1.4-point noise floor is retired here and left asserted in four files this PR does not touch.

- `CLAUDE.md:121` still instructs contributors to apply it
- `docs/architecture/PREDICTION_NOTES.md:78` still calls it a usable noise floor, directly contradicting the new 8.2
- `docs/architecture/FUZZY_RECOGNITION.md:258` reasons from it to justify a design decision
- `docs/research/PHILOSOPHY.md:123` quotes the whitepaper sentence this PR deletes

The last one has to change regardless, since it attributes a sentence to a document that will no longer contain it. The other three are a decision: either the convention survives as an explicitly-labelled working heuristic for deciding what to chase, which is defensible and worth keeping, or it goes everywhere.

Everything else in my earlier comment is optional polish, not a gate.

PR #133, opened 2026-09-16
fix(study): freeze the two learning paths on the keystroke path

owenpkent:
## Summary

The learning freeze was written from the methods with "learn" in the name, so it missed the two mutations that sit in `_press_char`: `observe_press` (every press carrying an offset) and `record_typed_word` (every completed word). `begin_study_capture` swaps only the synthesizer, so both ran on every keystroke of every trial.

- Gate both behind `_learning_frozen`.
- Replace the hand-written method list in the test file with an inventory that classifies every public method of `HybridPredictor` and fails on any it has not been told about.
- Correct the freeze comment and the CLAUDE.md entry, which both claimed the cover was complete.

## Why `observe_press` is the one that matters

The pointer bias it accumulates is read straight back by `FuzzyRecognizer.positions_for` on the next prediction. So this was not residue left behind a session: it moved the pills **inside the block being measured**, and further the longer the block ran, which is the one error shape counterbalancing cannot spread out. That is precisely the confound `docs/research/STUDY_PROTOCOL.md` section 5.2 says the freeze exists to prevent.

`record_typed_word` could put a word the participant had removed back on the bar part way through a condition, and leave it unblacklisted afterwards.

## Measurements

Replaying one block's worth of presses inside `frozen_learning()`, with the click offsets held identical between the two passes so the learned bias is the only thing that differs, over a 6,084 prefix sweep:

| | pointer slots learned under the freeze | top-5 set changed | top pill changed |
|---|---:|---:|---:|
| before | 26 | 88.2% | 44.0% |
| after | 0 | 0.0% | 0.0% |

Worth reading with a caveat: the flips concentrate on mis-click and fuzzy-rescue prefixes, where the spatial position dominates the ranking. Clean prefixes such as `hel` were stable throughout. So this is not 44% of real typing. It is the mis-clicked fraction of it, which is the fraction this study population produces most.

## The test is the durable part

Every existing test in `tests/test_learning_freeze.py` names one method, which is exactly how two keystroke-path mutations shipped unguarded: a suite that only checks the methods somebody remembered cannot report the one they forgot.

The file now partitions **every** public method of `HybridPredictor` into `_MUST_FREEZE`, `_LEAKS_PENDING_DECISION` or `_NOT_LEARNING_STATE`, and `test_every_public_method_is_classified` fails on anything unclassified, so a newly added mutation is a failing test until somebody decides which bucket it belongs in. Each mutating method needs a recipe, and each recipe is driven twice: frozen (must move nothing) and thawed (must move something), the second so a recipe that had quietly stopped exercising anything cannot satisfy the first vacuously.

Both failure modes were verified by reinstating them, not assumed:

- reinstating the `observe_press` bug fails with `observe_press wrote ['pointer'] while learning was frozen`
- adding an unclassified public method fails with a message naming it and the three buckets

Writing the inverse half also caught four of my own recipes that moved nothing: the 3-sighting candidate gate means a single `learn()` never reaches `user_vocab`, the token store's dict is `tokens` rather than `_tokens`, and `iphone` already ships in the proper-nouns table so both capitalisation recipes were writing a value that was already there.

## Deliberately not in this PR

Six **explicit user actions** still write through the freeze, and are pinned by `TestTheOutstandingLeaksAreStated` rather than gated: `blacklist_word`, `unblacklist_word`, `mark_bad_suggestion`, `remove_dispreference`, `unprefer`, `clear_user_data`.

Gating them silently is not obviously right. A control that quietly does nothing reads as a click that failed to register, which this project treats as the worse failure, so blocking them in the UI for the duration of a session is the likelier answer. That decision is not taken here. The test pins the current behaviour so it is stated rather than merely absent, and fails if one is gated, which is the prompt to move it into `_MUST_FREEZE`.

Also worth knowing and not addressed here: `StudyBridge.withdraw()` thaws but snapshots nothing, so the freeze is prevention only and anything that does slip past it is permanent. That is why the inventory test matters more than the guard list does.

## Related issue

No linked issue. Follows the isolation gap recorded in #132.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Test plan

- [x] Full `check.py` gate: ruff, format, mypy under both `--platform linux` and `--platform win32`, and the whole pytest suite. All green.
- [x] Added the inventory test described above, and verified it fails against both the reinstated bug and a new unclassified method.
- [x] Targeted run of the freeze, pointer, click-position and study suites (229 tests).
- [x] Before/after contamination sweep reported in the table above.

## Accessibility check

No interaction, timing or visual change. The two gated calls are inert only while a study session holds the freeze; ordinary typing is unaffected, so pointer-bias learning and blacklist rehabilitation behave exactly as before outside a session.

## Notes for reviewers

This changes the prediction engine's learning paths, so it is worth checking that the guards cannot fire outside a study session: `_learning_frozen` is set only by `frozen_learning()`, which only `StudyBridge` opens.

The interesting review question is not the two guards, it is whether the three buckets in the test file are classified correctly, and whether `_LEAKS_PENDING_DECISION` is the right call for the six explicit actions rather than gating them too.


tests/test_learning_freeze.py
```diff
@@ -270,3 +270,281 @@ def test_predict_still_returns_suggestions_while_frozen(
...
+        "user_total": ngram._user_total,
+        "bigrams": {k: dict(v) for k, v in ngram.bigrams.items()},
+        "trigrams": {k: dict(v) for k, v in ngram.trigrams.items()},
+        "capitalization": dict(ngram.capitalization),
+        "blacklist": set(ngram.blacklist),
+        "dispreference": dict(ngram.dispreference),
+        "preferred": dict(ngram.preferred),
+        "pointer": {k: tuple(v) for k, v in predictor._fuzzy.pointer._slots.items()},
+        "tokens": dict(ngram.tokens.tokens),
+    }
```
owenpkent, line 333: **Test coverage: include the learned state that can change independently of these fields.**

`HybridPredictor.learn()` also trains `_ppm` and `_ppm_word`, and `save()` persists the PPM model, but this snapshot contains no PPM state. I checked this with a runtime mutation at this PR's commit: insert `self._ppm.learn_text(text)` immediately before `learn()`'s frozen early return, then replay the new `learn` recipe. The call still returns `[]` and `_snapshot(before) == _snapshot(after)`, while `_ppm.total_chars` increases from 12939 to 12946. The frozen recipe therefore accepts a real model write. The thawed inverse does not catch it, because the normal unfrozen call still changes the n-gram fields.

There are similar omissions on the n-gram side: `_candidate_counts`, `_candidate_last_seen`, `_blacklist_type_count`, `taught_capitalization`, and the persisted `_user_bigrams`/`_user_trigrams` are absent. In particular, a sub-threshold candidate sighting or rehabilitation count can leave all displayed vocabulary/blacklist fields unchanged while carrying study residue into later typing.

Please extend the snapshot to cover all learned stores, preferably sharing their persistence representation where practical and including relevant in-memory learning state. Add recipe variants for an unknown word below promotion and a blacklisted word below rehabilitation, alongside the current cases. The current guards themselves are correct; this is a false-negative hole in the new regression protection.

tests/test_learning_freeze.py
```diff
@@ -270,3 +270,281 @@ def test_predict_still_returns_suggestions_while_frozen(
...
+        "get_capitalized",
+        "get_enabled_packs",
+        "get_key_alternatives",
+        "get_stats",
+        "get_unigram_freqs",
+        "get_user_packs_dir",
+        "import_vocabulary_pack",
+        "llmAvailableChanged",
+        "load_corpus",
+        "load_ppm_training_text",
```
owenpkent: **Inventory classification: corpus training is a learned-state mutation.**

These methods do not satisfy `_NOT_LEARNING_STATE`'s stated contract. `load_corpus()` calls `NgramPredictor.load_corpus()`, which calls `learn(text, corpus=True)`. That flag changes where context edges go; it still increments `user_vocab`, `_user_total`, candidate state and the decay counter. The hybrid method then trains PPM and calls `save()`. `load_ppm_training_text()` likewise trains the PPM trie. Neither checks the freeze.

For example, calling `load_corpus()` on a file containing `hello` inside `frozen_learning()` changes and saves the user's word count, yet the new inventory considers that method nonmutating and requires no recipe. `reload_from_disk` below also replaces learned state and is reachable from the Data Backup import flow. `enable_vocabulary_pack` above changes the n-gram tables through `apply_to_predictor`.

Please classify these as mutating maintenance/explicit-action APIs, with recipes, rather than grouping them with readers. They can go into a documented exception bucket if freezing them is outside this PR's scope; no silent UI no-op is required to fix the classification. The corpus methods have no current production callers, so this is not an additional ordinary-keystroke leak, but leaving them here undermines the inventory's central claim that every learned-state mutation has been identified and probed.

owenpkent (commented):
Reviewed commit `d5adfe7`.

The two production guards are correctly placed before their writes. `observe_press` no longer changes pointer state during the freeze, and `record_typed_word` stops both the rehabilitation counter and the eventual blacklist removal, returning the existing no-rehabilitation value. I found no additional implicit learned-state mutation in the normal prediction/autocorrect reader paths.

I ran the freeze, pointer-model, click-position (Python and QML), study capture, study session, study config, study metrics, and QML study-window suites against this exact commit in an isolated temporary tree with Qt offscreen. The test process exited successfully. I did not independently rerun the full `check.py` gate or reproduce the PR's 6,084-prefix measurement.

The two inline comments concern the new inventory's regression protection: the snapshot misses learned stores, and the non-learning bucket includes methods that train or replace those stores. These are test/classification issues, not evidence that either new production guard is broken.

One additional, pre-existing follow-up for the explicit-action decision: `KeyboardBridge.forgetToken` directly calls `self._predictor._ngram.tokens.forget(token)` (`src/keyboard_bridge.py:5665-5673`), and the dashboard invokes it. It therefore bypasses both `frozen_learning()` and the public-method inventory. The six named predictor methods are not the complete set of user-visible mutation paths. When deciding how the study UI handles such actions, include token deletion and Data Backup import/reload, and consider routing token deletion through a classified predictor API. Deferring the UI decision is reasonable for this focused fix, but the documented inventory boundary should remain explicit.

tests/test_learning_freeze.py
owenpkent: Confirmed and fixed in 90cedff, including by reproducing your exact mutation: putting `self._ppm.learn_text(text)` above `learn()`'s frozen early return left every assertion green before, and now fails with `learn:below promotion wrote ['ppm_total_chars'] while learning was frozen`.

You were right about the shape of the hole as well as the instance. A single thawed `learn()` of an unknown word moves `_candidate_counts`, `_candidate_last_seen` and `ppm_total_chars` and **nothing** the old snapshot looked at, so the whole sub-threshold band was invisible: two sightings of an unknown word, two of a blacklisted one.

The snapshot now covers `_candidate_counts`, `_candidate_last_seen`, `_blacklist_type_count`, `_user_bigrams`, `_user_trigrams`, `_ppm` and `_ppm_word`, and there are two new recipes (`learn:below promotion`, `record_typed_word:below rehabilitation`) that sit deliberately under each threshold. `taught_capitalization` is the one name from your list I did not add: it is not an attribute on `NgramPredictor` on this branch, so if it is coming from elsewhere I would rather add it in the same change as whatever introduces it than guess at the spelling.

tests/test_learning_freeze.py
owenpkent: Confirmed and fixed in 90cedff. I probed all four under the freeze before changing anything:

```
load_corpus              MUTATES   unigrams, user_vocab, user_total, bigrams, trigrams (+ ppm)
load_ppm_training_text   MUTATES   ppm
reload_from_disk         inert     (my probe had nothing on disk to reload, so this
                                    is a weak probe rather than a contradiction)
```

So you are right that calling them non-mutating made the inventory's central claim false. They now sit in `_MUTATES_OUTSIDE_THE_FREEZE`, documented as administrative rather than learning, with the reasons you gave: none is on the keystroke path and the corpus loaders have no production caller.

I took your "documented exception bucket" option rather than freezing them, since freezing is the same open question as the explicit actions. Two things to stop the bucket becoming a parking space: `test_the_unfrozen_maintenance_apis_really_do_mutate` drives the two that are drivable and asserts they do write, and the five needing a pack directory or a written model file are named in `_NO_RECIPE_NEEDS_FIXTURE` alongside the suites that do cover them, so "classified but unprobed" is explicit rather than implied by bucket membership.

On `forgetToken` from your top-level review: agreed, and it is a better example than any of mine because it bypasses `HybridPredictor` entirely (`self._predictor._ngram.tokens.forget(token)`), so no inventory over that class's public API can see it. The Debug panel's `importTextFile` calls `self._predictor._ngram.learn(text)` the same way. I have left both out of this PR deliberately: the inventory's boundary is `HybridPredictor`, and widening it to "every bridge slot that reaches a store" is the same decision as the explicit actions rather than a test change. Worth folding into that decision, with routing token deletion through a classified predictor API as you suggest.

tests/test_learning_freeze.py
owenpkent: Correction to my reply above: I was wrong that `taught_capitalization` does not exist. It is `NgramPredictor.taught_capitalization`, a `set[str]` at ngram_predictor.py:147, separate from the `capitalization` dict, so a write that only marked a word as user-taught moved nothing the snapshot looked at. Added in 8678424. Your list was complete; my check for it was not.

PR #137, opened 2026-09-17
docs: record the shared caches and the 1.5.0 follow-ups

owenpkent:
## Summary

Documentation catch-up after 1.5.0 shipped. No code changes.

- **CLAUDE.md**: document the two process-level caches, which were the half of 1.5.0 that made the expanded vocabulary affordable and whose reasoning was nowhere.
- **TODO.md**: the two follow-ups from the pre-release review.
- **Test count**: corrected in three files, where it had drifted 28%.

## Why the caches needed writing down

They share state between predictor instances, which is the highest-risk shape of change in this engine, and three things about them are not obvious from reading the code:

- Sharing is safe **only** because the packed representation is immutable: every method on `PackedDeletes` and `PackedPrefixes` past `__init__` is a read accessor, and everything that changes as the user types goes to the mutable overlay beside them.
- The digest **must** be collision-resistant. Python's `hash` is not, and a collision here would not raise: it would quietly answer one vocabulary's predictions from another vocabulary's index.
- The taught-acronym guard reads **`taught_capitalization`**, not `capitalization`. Proper nouns fill the latter and leave the former empty, so guarding on the wrong one bypasses the cache on every construction and the caching silently does nothing. That is how the first version shipped, and it is worth a paragraph because the two call paths already disagree about when proper nouns load.

## The two follow-ups

1. **`explicit_filter_available` is not wired to the UI.** `_load_explicit_words` fails open by design, so a missing word list leaves the Filter Explicit Words toggle reading as on while it suppresses nothing, with only a log line to say so. `passwordDetectionAvailable` is the pattern to copy.
2. **Cold start is about 3 s against 0.8 s.** The caches cannot help the first predictor in a process by construction. Closing it means shipping the packed buffers as data rather than computing them at launch, which is real work and is the main measured cost of 1.5.0.

Recorded rather than left to be rediscovered as "the keyboard got slower to start".

## The test count

Claimed ~2,250, live count 2,888 across 82 files. CLAUDE.md asks for that number not to be restated elsewhere **precisely because it drifts**, and README states it twice with CONTRIBUTING once. All three are corrected here, but the restatement is the actual problem and is worth revisiting: the rule and the practice disagree.

## Related issue

No linked issue. Follows the 1.5.0 release.

## Type of change

- [x] Docs only

## Test plan

- [x] Full `check.py` gate (nothing should move, and nothing did).
- [x] Test count taken from `pytest --collect-only -q` rather than estimated.
- [x] Checked the diff introduces no em dashes.

## Accessibility check

N/a: documentation only.

## Notes for reviewers

The claim most worth checking is the taught-acronym paragraph, since it describes a bug that existed for one commit and the fix depends on which of two similarly-named tables is read.


owenpkent (commented):
Reviewed `0013ec3` against its base and checked the cache claims against `packed_cache.py` and `NgramPredictor._validated_extra_words`. The BLAKE2b key, two-entry limit, mutable overlays, and `taught_capitalization` bypass match the implementation. No blocking documentation defect found.

Two integration notes: once #143 lands, put this new cache explanation in `docs/architecture/PREDICTION_NOTES.md` and leave a short pointer in `CLAUDE.md`. Also consider replacing the repeated README/CONTRIBUTING test counts with the collection command rather than updating another set of numbers that will drift. These are nonblocking maintenance suggestions. No local test run was needed for this documentation-only diff; its CI run passed.

PR #138, opened 2026-09-19
fix: relauncher survives updates; shortcuts carry the taskbar identity

owenpkent:
## Summary

Two production defects found from one machine's evidence (`relauncher.log` had no production entry ever; the installed exe's mtime predated the install that wrote it), plus the duplicate-taskbar-icon report.

**1. The installer was killing its own relauncher.** The post-update helper ran as a second `alpha-osk.exe`, so `customCloseAlphaOsk`'s `taskkill /IM` and its wait-for-exit loop spent the full ~8 s budget staring at it and then force-killed it, on every update. The splash never showed, the relaunch never ran, and the keyboard only came back via the `Exec explorer.exe` fallback. The helper now runs from a renamed copy of the bundle in `%TEMP%` (`alpha-osk-relauncher.exe`). Both halves are load-bearing: the different image name keeps it clear of the taskkill, and the different location keeps the install dir unlocked. Exempting the helper's PID from the kill was considered and rejected: a process running from the install dir holds its exe and DLLs mapped, NSIS cannot overwrite a mapped file, and a silent install that cannot write aborts, which strands the user with no keyboard at all. Stale staging copies are swept at the next spawn; dev runs no longer spawn a helper.

**2. The helper's readiness check was unsatisfiable.** It required the new exe's mtime to be newer than the old keyboard's death, and NSIS restores build-machine timestamps on extracted files (`SetDateSave` defaults on), so even a surviving helper would have polled 180 s and then claimed the update failed. The updater now snapshots the installed exe's mtime before the install (`--old-exe-mtime`) and the helper waits for it to differ. Without a snapshot it falls back to the legacy comparison. The splash also ignores WM_CLOSE (hide, keep polling) and sets quit-on-last-window-closed off, so only its own state machine can end the run.

**3. Duplicate taskbar icons.** The app has set an explicit AppUserModelID on itself since v1.2.0, but the installer's shortcuts never carried it, so Windows derived a different identity for a pin and showed two buttons for one app (verified on a real machine: a pre-v1.2.0 pin with an empty id beside the running window's explicit id). The Install section now stamps `System.AppUserModel.ID` onto the Start Menu and desktop shortcuts through the shell property store (`System::Call`, no new plugin), unconditionally, so shortcuts left behind by older installs are stamped on the next update too. `build.py` reads the id out of `keyboard_app.py` at generation time so the two cannot drift. The uninstall shortcut is deliberately not stamped.

All of it takes effect on the update *after* the release that ships it, the same "the fix cannot protect the upgrade that delivers it" pattern as the settings-preserving uninstall.

## Flagged per project rules

- **Build pipeline touched**: `build/windows/build.py` (new `StampShortcutAppId` function, `APP_AUMI` define read from `keyboard_app.py`, stamping calls in the Install section) and `build/windows/installer.nsh` (new `customStampShortcutAppId` macro). No signing, packaging, or release-flow changes; the stamping is best-effort and can never fail the install.
- Prediction engine and telemetry untouched.

## Type of change

- [x] Bug fix

## Test plan

- `python check.py` green (ruff, format, mypy linux + win32, full suite).
- New: `tests/test_updater.py::TestRelauncherSpawn` (helper name, staging location, mtime snapshot in argv, stale-stage sweep, non-fatal staging failure, console flags), `tests/test_update_relauncher.py::TestAFreshInstallIsSeenThroughItsBuildTimestamp` (the build-stamped-mtime production case end to end, unchanged-mtime inverse, no-snapshot fallback in both directions), `tests/test_windows_installer.py::TestTheShortcutsCarryTheTaskbarIdentity` (id matches `keyboard_app.py`, both shortcuts stamped, stamping survives unchecked boxes, property-store pieces, and a `makensis` compile of the macro, which ran rather than skipped locally).
- End-to-end verification of the full update flow needs the next signed release; the unit and compile coverage above is what can be checked before that.

## Accessibility check

The update gap is when the user has no keyboard at all. This PR makes the progress splash actually appear during it, removes ~8 s of dead wait from every update, and keeps the relaunch path that returns the keyboard without any manual step.


build/windows/installer.nsh
```diff
@@ -168,6 +168,58 @@
...
+  Push $2
+  Push $3
+  Push $4
+  Push $5
+  ${If} ${FileExists} "$0"
+    System::Call 'shell32::SHGetPropertyStoreFromParsingName(w r0, p 0, i 2, g "{886d8eeb-8cf2-4446-8d02-cdba1dbdcf99}", *p .r1) i .r2'
+    ${If} $2 = 0
+      System::Call '*(g "{9F4C2855-9F79-4B39-A8D0-E1D42DE1D5F3}", i 5) p .r3'
+      System::Call '*(w "${APP_AUMI}") p .r4'
+      System::Call '*(&i2 31, &i2 0, &i2 0, &i2 0, p r4) p .r5'
```
owenpkent, line 206: [P2] Allocate the characters rather than a structure containing their pointer

`*(w "${APP_AUMI}")` allocates a structure containing a string pointer. `$4` is the address of that structure, so putting `p r4` into the VT_LPWSTR payload makes `SetValue` read the pointer bytes as UTF-16 text. I compiled and ran this macro on a temporary `.lnk`: `GetValue(PKEY_AppUserModel_ID)` returned VT_LPWSTR with garbage instead of `OKStudio.AlphaOSK`, despite successful HRESULTs. Thus newly created and updated shortcuts still cannot share the app's taskbar identity.

Allocate an inline WCHAR buffer (`&wN`, with room for the terminator) and pass its address here. Changing the allocation makes the readback match the intended ID. Please cover a real temporary shortcut with write/readback, since merely compiling the macro does not exercise these runtime layouts. See the [NSIS System parameter types](https://nsis.sourceforge.io/Docs/System/System.html).

owenpkent (commented):
Found one correctness issue in shortcut stamping, detailed inline: the macro stores a pointer-to-pointer where VT_LPWSTR expects a pointer to UTF-16 characters, so the persisted AppUserModelID does not match the running application.

Validation at `c54a9ab`: all 212 updater, relauncher, and Windows-installer tests passed. I also compiled and executed a minimal, unelevated NSIS harness against a temporary shortcut: the original macro committed an invalid ID; a character-buffer allocation produced `OKStudio.AlphaOSK`. The current compile/text assertions miss this failure, so add a runtime write/readback regression test. The full signed update handoff was not exercised.

PR #139, opened 2026-09-19
fix(compact): balance the emphasis on Backspace and Enter

owenpkent:
## Summary

- **Backspace stops out-shouting Enter on the compact grid.** The two keys that destroy text (Backspace, Del) now take the plain border instead of the full-strength accent ring. The fills were never the cause and do not move: on Dark they sit 13.8 and 11.8 OKLab dE from a plain key, which is level. It was the ring, four of them against Enter's nought.
- **Enter's `#2a5a2a` is gone.** It was the only fill in `Main.qml` that skipped `washFor`'s contrast walk, and the only literal hue left in a project whose first colour rule is that there are none. It measured **1.89:1 on Typewriter, 2.15 on Light and 3.28 on Vaporwave**, all under WCAG AA. Enter now shares the editing keys' wash, so the `off` scheme carries one hue rather than two and the ratio is fixed for free.
- Docs updated in the same sitting: the CLAUDE.md compact-view bullet and `docs/architecture/COMPACT_VIEW.md` both stated the old rule as authoritative.

## Related issue

No issue, reported directly.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs only
- [ ] Build / CI / tooling
- [ ] Other (describe)

## Test plan

Four new cases for the border split (`TestTheKeysThatDestroyTextTakeNoRing`) and four for Enter's fill (`TestEnterSharesTheEditingKeysWash`), all measured off the **live** `borderColor` / `_roleFill` on rendered `KeyButton`s rather than recomputed from the properties the QML sets, and swept across all nine themes.

Every one was mutation-checked, since a colour assertion that cannot fail is easy to write here:

| Mutation | Caught by |
|---|---|
| Restore the ring on all five accent keys | `test_backspace_and_delete_carry_the_plain_border`, `test_backspace_no_longer_outranks_enter`, `test_it_holds_on_every_theme` |
| Drop the ring from every key | `test_the_other_accent_keys_still_ring`, `test_it_holds_on_every_theme` |
| Bring back the `#2a5a2a` literal | `test_enter_is_painted_with_the_editing_keys_wash`, `test_the_fill_is_theme_derived_and_legible_on_every_theme` |
| Mark Enter `style: "accent"` in the layout JSON | `test_enter_did_not_become_an_accent_key` |

That last one is the trap worth knowing about: it paints Enter identically and hands it the ring back, so it satisfies the fill assertion while quietly being a different decision.

- [x] `python check.py` passes locally — with a caveat, see **Notes for reviewers**
- [x] Added or updated tests
- [x] Tested in `python run.py` (not just frozen build) — rendered headlessly from the real `Main.qml`, both view modes, all nine themes, both `mono` and `off`
- [x] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

## Accessibility check

Nothing about hit targets, timing or repeat changes. The contrast direction is strictly better: the worst Enter label ratio across the nine themes goes from **1.89:1 to 4.51:1**, and Backspace keeps its wash, so it is still marked, just no longer the loudest thing on the board.

## Notes for reviewers

**The border is the one colour on a key that Key Colours never reaches.** It keys off the layout JSON's `style` while every fill keys off `role`. That is why the ring won even on Monochrome, whose stated intent in `palette.js` is to make Enter the brightest key on the board. Making the border follow the scheme properly (`kill` red, `commit` green) was the larger option on the table and was not taken; it is written up in both docs as still available.

**Two things that look like harmless simplifications and are not**, both called out in the code comments and pinned by tests:

1. Naming `backspace` / `delete` directly in `keyBorderFor` instead of asking `Palette.roleForKey`. That function is already the project's one answer to the question, and a second list in `Main.qml` is the pair of parallel blocks that drift.
2. Marking Enter `style: "accent"` to collapse the switch further. Paints the same, rings differently.

**Scope of the Enter change:** only the `off` scheme reaches that line. Every other scheme resolves Enter through `_roleFill`, so under Monochrome (the shipped default) there was never any green and Enter was already the brightest key. That is why the contrast failure went unnoticed. The accepted cost of the fix is that Enter and Backspace are identical under `off`, chosen over three alternatives that each spent more colour rather than less (a muted commit-hue wash; keeping the green and flipping the label white; the commit hue at full chroma, which additionally needs a per-key ink override to clear Vaporwave and so new plumbing on `NumpadPanel`).

**On the local gate:** `ruff`, `ruff format` and both `mypy` passes are clean, and the QML suites pass serially. `pytest -n auto` crashes an xdist worker on this machine with a `Windows fatal exception: access violation`, which then fails whichever test that worker was running. **It is not from this change** — I stashed the branch and reproduced it on a clean tree, where it took down `tests/test_qml_window_menu.py` instead, and a `-n 4` run crashed a different worker on `tests/test_windows_window.py`. A serial run is the trustworthy signal here; CI's own sharding is the authority. Worth a separate look at some point.


owenpkent (commented):
Reviewed `57e8f5b`. No blocking issue found in the production changes. The border decision reuses the existing `kill` role, Enter keeps its separate style, and both the grid and numpad now receive the theme-derived wash when the role-colour scheme is off.

Ran the eight new border/fill tests at this commit: all passed, including the theme sweeps. This is targeted validation, not a rerun of the full suite. When integrating with #143, keep the detailed colour rationale in the architecture docs so the onboarding split is preserved.

PR #140, opened 2026-09-19
feat(window): snap the keyboard to screen edges while moving it

owenpkent:
## Summary

- Moving the keyboard window now snaps it flush to a screen edge, or to the screen's horizontal centre, once the proposed position comes within 24 px. Both ways of moving it go through one function, `Main.qml::snapWindowPos`: the title-bar drag and the click-free Move mode.
- Landing a drag flush against an edge otherwise means holding the button down for the whole travel and then releasing within a pixel or two of the right spot, which is the gesture this keyboard exists to avoid needing.
- New toggle: *Settings -> Appearance -> **Window** -> "Snap to Screen Edges"*, default ON.

## Related issue

No issue; asked for directly.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs only
- [ ] Build / CI / tooling
- [ ] Other (describe)

## Test plan

- `python check.py` (ruff, ruff format, mypy under both `--platform linux` and `--platform win32`, full pytest) run on this branch.
- 9 new headless QML tests in `tests/test_qml_window_menu.py::TestSnappingToScreenEdges`, driving real `QTest` mouse events rather than the QML API, so a version where `dragArea` never called the snap would fail rather than pass quietly.
- **Falsifiability checked by mutation.** Five defects were reinstated one at a time and each failed exactly the guard that owns it:
  | reinstated defect | test that caught it |
  |---|---|
  | Move mode accumulates into `root.x` (the edge becomes a trap) | `test_move_mode_can_leave_an_edge_it_snapped_to` |
  | No re-anchor by the applied motion (same travel counted twice) | `test_move_mode_holds_an_edge_until_the_pointer_has_really_left_it` |
  | The drag path never calls `snapWindowPos` | `test_a_title_bar_drag_lands_flush_on_an_edge` |
  | Snap clamps instead of attracting (no threshold) | `test_a_position_past_the_zone_is_left_exactly_where_it_is` |
  | Vertical centre made a target too | `test_the_horizontal_centre_is_a_target_and_the_vertical_one_is_not` |
- Re-ran every suite that loads `Main.qml` (`test_qml_window_menu`, `test_qml_compact_view`, `test_qml_panel_placement`, `test_qml_key_colors`, `test_qml_prediction_bar`, `test_window_corners`).

- [x] Added or updated tests
- [ ] Verified on Windows
- [ ] Verified on Linux
- [ ] Verified on macOS

**Not yet driven by hand.** Whether 24 px is the right pull is a judgement only a real pointer can make, and the headless plugin gives one screen, so the multi-monitor half is unexercised (see below). Worth a manual pass on Windows before merge.

## Accessibility check

- [x] Works with default repeat-delay / repeat-interval settings (nothing on the keystroke path changed)
- [x] No new requirement for fast or precise input. This is the opposite: it *removes* the requirement that a drag be released within a pixel or two of an edge to sit flush against it. 24 px rather than the ~10 a mouse-driven desktop uses, for the same reason `hitMarginH` exists.
- [x] Visual feedback still works for low-vision users (no colour, contrast or layout change)

The setting is there because snapping is not free: it costs the ability to park the window a few pixels off an edge deliberately. Default ON because for this pointer the trade lands clearly one way.

## Notes for reviewers

**The one thing worth reading closely is why the snapped value is never written back into whatever the caller accumulates.** Feed it back and every later delta is measured from the snap point, so a pointer moving inside the zone can never build up the travel it needs to leave, and the edge holds the window for good. `dragArea` is immune for free, because it recomputes its proposal from the press origin on every event. `windowMoveOverlay` is not: its whole design is accumulated deltas in its own coordinates, so it now keeps an unsnapped shadow position (`freeX` / `freeY`) and shows the snapped version of it.

**Move mode also has to re-anchor by however far the window *actually* went** (`anchorX = mouse.x - (root.x - beforeX)`). The self-correction the mode rests on assumes the window moved the whole delta, which slides it out from under the pointer and puts the pointer back on the anchor. While a snap holds the window still it does not, so measuring the next event against the old anchor counts the same travel again on every event and the pointer leaves the zone in a fraction of the threshold. Compensating keeps `anchor == pointer - window` true either way. This is why the escape test walks out in three 10 px steps rather than one 30 px jump: a single jump passes against the re-anchoring bug as well.

**Deliberately not done:**
- Snapping is *not* extended to the Snippets and Symbols windows. Those get dragged clear of the field being filled in, which is a "not here" gesture rather than an "exactly there" one, and they have no Move mode to share the shadow-position machinery with.
- The left/right resize handles are untouched.
- Horizontal centre is a target and vertical centre is not. Centring a wide, short keyboard across the screen is something people do; parking it halfway down is not, and a snap nobody wanted reads as the window sticking for no reason.

**Known coverage gap, stated in the test class docstring rather than papered over:** the offscreen plugin gives exactly one screen, so nothing headless can tell a `screenBoundsAt` lookup from a primary-screen one. Every target in the tests is therefore derived from `screenBoundsAt` rather than `Screen.width`, which is the same limitation `tests/test_qml_panel_placement.py` carries and the same reason. The multi-monitor case matters here: a monitor to the left has negative coordinates a primary-screen calculation cannot express, which is the bug the snippets restore documents one window over.

**One test-file change beyond the new class:** `_park` now turns snapping off. That is not tidiness. Every displacement assertion in that file reads "the window followed the pointer by exactly the delta", which is only true when no edge intervened, and whether one does depends on a screen size the offscreen plugin chooses for us. Stating the precondition beats a suite that passes because 200 px happened to be more than `snapThreshold` from everything.

No change to the prediction engine, the build/signing pipeline, or telemetry.


owenpkent (commented):
Reviewed `7c85dca` against the actual base, `docs/slim-claude-md`. No blocking defect found in the move logic. The drag path recomputes from the press origin, while Move mode preserves the unsnapped position and compensates the anchor by actual window movement; that avoids both a trapped edge and double-counted pointer travel. The setting reaches both paths.

All 24 window-menu tests passed at this commit. Remaining coverage is the stated multi-monitor gap: the tests use one offscreen display. A useful follow-up is to substitute synthetic screen bounds and cover a monitor at negative coordinates plus a transition between displays. A desktop pass should also check bottom-edge placement with a visible taskbar and mixed DPI.

